PR #12, opened 2017-08-11
Change store.menu reference to "block"

jahvi:
The `store.menu` element [is a block](https://github.com/magento/magento2/blob/99e85cbc45223baa3551e4c534b650c0d2c6358b/app/code/Magento/Theme/view/frontend/layout/default.xml#L65), just tweaking the example to reflect that.


Igloczek:
Thanks for the contribution!

PR #15, opened 2017-09-14
Remove php7 type hints

vlradstake:


DariuszMaciejewski:
`array` is a type hint added in PHP 5.1 (not PHP 7), so we don't want to remove it.

Issue #23, opened 2017-11-09
Module does not work when database tables have prefixes


alexvandervegt:
The module does not work when you have table prefixes like:

Exception #0 (Zend_Db_Statement_Exception): SQLSTATE[42S02]: Base table or view not found: 1146 Table 'databasename.eav_attribute' doesn't exist, query was: SELECT `a`.`attribute_id` FROM `eav_attribute` AS `a`

This should be:
databasename.eav_attribute => databasename.prefix_eav_attribute

DariuszMaciejewski:
@alexvandervegt  thank you for your report.

DariuszMaciejewski:
The issue has been fixed and delivered to the develop branch.

PR #28, opened 2017-11-19
Fix sql queries when table prefix is set in Magento

DariuszMaciejewski:
Related to the issue #23


DariuszMaciejewski:
The issue has been fixed and delivered to the develop branch.

PR #37, opened 2017-12-18
Set block node classes from node model (closes #36)

### LINKED ISSUE #36: Cann't access node classes from template
```phtml

<?php /** @var \Snowdog\Menu\Block\NodeType\Category $block */ ?>

$class = $block->getMenuClass();
```

I can call `getMenuClass` but there is no method `getNodeClasses`, so node css classes attribute looks like useless.

Koc:
Fixes #36


Igloczek:
Internal ticket DEV-37355

PR #39, opened 2018-01-11
Problem with overwrite node template if it's more than one menu

dawiddczaja:
Internal ticket DEV-37921

### DIFF
--- a/Block/NodeType/AbstractNode.php
+++ b/Block/NodeType/AbstractNode.php
@@ -136,15 +136,7 @@ abstract class AbstractNode extends Template implements NodeTypeInterface
     {
         return $this->viewAllLink;
     }
-
-    /**
-     * @return string
-     */
-    protected function getDefaultTemplate()
-    {
-        return '';
-    }
-
+    
     /**
      * @return string
      */
--- a/Block/NodeType/AbstractNode.php
+++ b/Block/NodeType/AbstractNode.php
@@ -136,7 +136,7 @@ abstract class AbstractNode extends Template implements NodeTypeInterface
     {
         return $this->viewAllLink;
     }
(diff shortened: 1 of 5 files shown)


DariuszMaciejewski:
@dawiddczaja could you explain why we can't use `$this->_template`?

dawiddczaja:
@DariuszMaciejewski Yes sure, when node (e.g. category) is rendered first time `$this->_template` was set in property like  `protected $_template = 'menu/node_type/category.phtml';` then it's go to the function 

https://github.com/SnowdogApps/magento2-menu/blob/master/Block/NodeType/AbstractNode.php#L138

And if template is different then default,  function ` $this->setTemplate($template);` assigned to variable `$this->_template` new value so next time when same node try render it's paste argument to function `getMenuTemplate` it's not default template.

This behavior corrupts function https://github.com/SnowdogApps/magento2-menu/blob/master/Model/TemplateResolver.php#L36

Block/NodeType/AbstractNode.php
```diff
@@ -131,7 +136,7 @@ public function isViewAllLinkAllowed()
     {
         return $this->viewAllLink;
     }
-
+
```
DariuszMaciejewski: remove extra whitespace

Block/NodeType/AbstractNode.php
```diff
@@ -21,6 +21,11 @@
     const NAME_CODE = 'node_name';
     const CLASSES_CODE = 'node_classes';
 
+    /**
+     * @var string
+     */
+    protected $defaultTemplate = '';
```
maciekkossuda: Empty string is not the best solution to initialize var.

PR #43, opened 2018-01-29
Fixed compatibility with EE 2.2

DariuszMaciejewski:
Internal ticket DEV-38534


Helper/EavStructureWrapper.php
```diff
@@ -2,219 +2,124 @@
...
-        }
-    }
+    private $categoryIdentifierField;
 
     /**
-     * @return string
+     * @param MetadataPool $metadataPool
      */
-    public function getEdition()
+    public function __construct(MetadataPool $metadataPool)
```
KamilBalwierz: `Magento\Framework\EntityManager\MetadataPool` is available since Magento 2.1 - do we drop support for Magento 2.0 ? (also dependency on framework in at least 100.1.0 should be placed in composer.json)

If we are using `MetadataPool` I see no point of using and maintaining `EavStructureWrapper` class as it only acts as proxy for `MetadataPool` (I would use `MetadataPool` directly in resource models)

DariuszMaciejewski: Okay, I removed `EavStructureWrapper` and I updated composer dependencies.

Moreover, we decided to release a new minor version and drop support for old core versions.

PR #49, opened 2018-03-19
#38867 Update target attribute

dawiddczaja:


Setup/UpgradeSchema.php
```diff
@@ -91,4 +95,29 @@ private function addTargetAttribute(SchemaSetupInterface $setup)
...
+        $connection->update(
+            $table,
+            ['target' => 1],
+            "target = '_blank'"
+        );
+        $connection->modifyColumn(
+            $table,
+            'target',
+            [
+                'type' => Table::TYPE_SMALLINT,
```
maciekkossuda: You can use `TYPE_BOOLEAN` here

PR #50, opened 2018-04-05
Feature/39120

dawiddczaja:
[DEV-39120]


Block/NodeType/Product.php
```diff
@@ -0,0 +1,228 @@
...
+    public function fetchData(array $nodes)
+    {
+        $storeId = $this->_storeManager->getStore()->getId();
+
+        list(
+            $this->nodes,
+            $this->productUrls,
+            $this->productPrices,
+            $this->productImages
+            ) = $this->productModel->fetchData($nodes, $storeId);
```
jochhop: I think moving the bracket to the same level as `list` will look better

Block/NodeType/Product.php
```diff
@@ -0,0 +1,228 @@
...
+     * @return false|string|double
+     */
+    public function getProductData($data, $nodeId)
+    {
+        if (!isset($this->nodes[$nodeId])) {
+            throw new \InvalidArgumentException('Invalid node identifier specified');
+        }
+
+        $node = $this->nodes[$nodeId];
+        $productId = (int) $node->getContent();
```
jochhop: Please remove the space between the casting and the variable

Model/NodeType/Product.php
```diff
@@ -0,0 +1,105 @@
...
+        $productIds = [];
+
+        $websiteId = $this->storeManager->getStore()->getWebsiteId();
+        $customerGroupId = $this->customerSession->getCustomer()->getGroupId();
+
+        foreach ($nodes as $node) {
+            $localNodes[$node->getId()] = $node;
+            $productIds[] = (int)$node->getContent();
+        }
+
```
jochhop: There is an extra enter here

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        $connection = $this->getConnection('read');
+        $table = $this->getTable('url_rewrite');
+        $select = $connection
+            ->select()
+            ->from($table, ['entity_id', 'request_path'])
+            ->where('entity_type = ?', 'product')
+            ->where('redirect_type = ?', 0)
+            ->where('store_id = ?', $storeId)
+            ->where('entity_id IN (' . implode(',', $productIds) . ')');
```
jochhop: As in the previous lines you use `?` to add values in where clause, I think we can use it here also so we can keep the same structure (it can be applied to the below methods also)

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,129 @@
...
+    private function getAttributeIdByCode($connection, $code)
+    {
+        $select = $connection->select()->from(
+            ['a' => $this->getTable('eav_attribute')],
+            ['attribute_id']
+        )->join(
+            ['t' => $this->getTable('eav_entity_type')],
+            't.entity_type_id = a.entity_type_id',
+            []
+        )->where('t.entity_type_code = ?', CoreProduct::ENTITY)->where(
```
jochhop: Please move the 2nd `where` to the next line (I think we can reformat all of this to keep the same structure as with `join`

Block/NodeType/Product.php
```diff
@@ -0,0 +1,228 @@
...
+     * @throws \InvalidArgumentException
+     */
+    public function isCurrentProduct($nodeId)
+    {
+        if (!isset($this->nodes[$nodeId])) {
+            throw new \InvalidArgumentException('Invalid node identifier specified');
+        }
+
+        $node = $this->nodes[$nodeId];
+        $productId = (int) $node->getContent();
```
jochhop: Please remove the space between casting and variable

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        $connection = $this->getConnection('read');
+        $table = $this->getTable('url_rewrite');
+        $select = $connection
+            ->select()
+            ->from($table, ['entity_id', 'request_path'])
+            ->where('entity_type = ?', 'product')
+            ->where('redirect_type = ?', 0)
+            ->where('store_id = ?', $storeId)
+            ->where('entity_id IN (?)', implode(',', $productIds));
```
jochhop: Sorry that I didn't spot it before, but I think this will work without `implode`, just using the array as parameter (please also check below methods, I can see this is repeated below)

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+
+        return $connection->fetchPairs($select);
+    }
+
+    /**
+     * @param AdapterInterface $connection
+     * @param string $code
+     * @return array|string
+     */
+    private function getAttributeIdByCode($connection, $code)
```
DariuszMaciejewski: Can't we use `\Magento\Eav\Model\Config::getAttribute()` ?

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+
+        $connection = $this->getConnection('read');
+        $nameAttributeId = $this->getAttributeIdByCode($connection, 'image');
+
+        $table = $this->getTable('catalog_product_entity_varchar');
+        $select = $connection->select()
+            ->from($table, [$linkField, 'value'])
+            ->where('attribute_id = ?', $nameAttributeId)
+            ->where('store_id = ?', 0)
+            ->where($linkField .' IN (?)', $productIds);
```
DariuszMaciejewski: `$linkField` can be `row_id` so this condition seems to be invalid. Can't we use product collection (I mean resource model) to load all things?

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+        $linkField = $metadata->getLinkField();
+
+        $connection = $this->getConnection('read');
+        $nameAttributeId = $this->getAttributeIdByCode($connection, 'image');
+
+        $table = $this->getTable('catalog_product_entity_varchar');
+        $select = $connection->select()
+            ->from($table, [$linkField, 'value'])
+            ->where('attribute_id = ?', $nameAttributeId)
+            ->where('store_id = ?', 0)
```
DariuszMaciejewski: Why can't we load data for the current store?

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+     * @return array
+     * @throws \Exception
+     */
+    public function fetchConfigData()
+    {
+        $metadata = $this->metadataPool->getMetadata(ProductInterface::class);
+        $identifierField = $metadata->getIdentifierField();
+        $connection = $this->getConnection('read');
+
+        $select = $connection->select()->from(
```
DariuszMaciejewski: This query can break the store if the catalog has a lot of products.

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+            ['e' => $this->getTable('catalog_product_entity')],
+            [$identifierField, 'sku']
```
DariuszMaciejewski: We can have many records with the same identifier (entity_id) in case of Magento Commerce.

view/frontend/templates/menu/node_type/product.phtml
```diff
@@ -0,0 +1,31 @@
...
+<?php
+$nodeId = $block->getId();
+$class = $block->getMenuClass();
+$productUrl = $block->getProductUrl($nodeId);
+$productPrice = $block->getProductPrice($nodeId);
+$productImage = $block->getProductImage($nodeId);
+$productUrl = $productUrl ?: '#';
+$dataAttribute = '';
+
+if ($block->getIsViewAllLink()) {
```
DariuszMaciejewski: Is this necessary for product nodes?

Model/ResourceModel/NodeType/Product.php
dawiddczaja: Yes we can but it's slower than this query and i think in menu module we should use the fastest solution

view/frontend/templates/menu/node_type/product.phtml
dawiddczaja: @talalus please check comment

Model/ResourceModel/NodeType/Product.php
dawiddczaja: Yes `$linkField` can be entity_id and row_id so I think it's correct because `catalog_product_entity_varchar` for EE have `row_id` and CE have `entity_id` also I used sql query because it's faster.

Model/ResourceModel/NodeType/Product.php
dawiddczaja: You're right fixed

Model/ResourceModel/NodeType/Product.php
dawiddczaja: My idea is add grid instead of search by name/sku/id or remove option autocomplete what You think ?

Model/ResourceModel/NodeType/Product.php
dawiddczaja: I think if we use one of my idea from prev comment this query will be unnecessary

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        $connection = $this->getConnection('read');
+        $table = $this->getTable('url_rewrite');
+        $select = $connection
+            ->select()
+            ->from($table, ['entity_id', 'request_path'])
+            ->where('entity_type = ?', 'product')
+            ->where('redirect_type = ?', 0)
+            ->where('store_id = ?', $storeId)
+            ->where('entity_id IN (?)', $productIds);
```
DariuszMaciejewski: Add `WHERE metadata IS NULL` to select url rewrites for product pages.

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,132 @@
...
+    public function fetchImageData($storeId = Store::DEFAULT_STORE_ID, $productIds = [])
+    {
+        $metadata = $this->metadataPool->getMetadata(ProductInterface::class);
+        $linkField = $metadata->getLinkField();
+
+        $connection = $this->getConnection('read');
+        $nameAttributeId = $this->getAttributeIdByCode($connection, 'image');
+
+        $table = $this->getTable('catalog_product_entity_varchar');
+        $select = $connection->select()
```
DariuszMaciejewski: You should use product collection (not Magento API because it is slow) to load this data. Your solution is incompatible with the staging module in Magento Commerce.

Model/ResourceModel/NodeType/Product.php
dawiddczaja: done

Model/ResourceModel/NodeType/Product.php
dawiddczaja: Removed

Model/ResourceModel/NodeType/Product.php
dawiddczaja: autocomplete functionality removed

Model/ResourceModel/NodeType/Product.php
dawiddczaja: done

Model/ResourceModel/NodeType/Product.php
dawiddczaja: After changes function is unnecessary (removed)

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,100 @@
...
+     */
+    private $productCollection;
+
+    public function __construct(
+        ResourceConnection $resource,
+        MetadataPool $metadataPool,
+        Collection $productCollection
+
+    ) {
+        $this->metadataPool = $metadataPool;
```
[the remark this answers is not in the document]
dawiddczaja: removed

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * @var Collection
+     */
+    private $productCollection;
+
+    public function __construct(
+        ResourceConnection $resource,
+        MetadataPool $metadataPool,
+        Collection $productCollection
+
```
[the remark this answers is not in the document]
dawiddczaja: removed

Model/ResourceModel/NodeType/Product.php
```diff
@@ -0,0 +1,100 @@
...
+            ->where('entity_id IN (?)', $productIds);
+
+        return $connection->fetchPairs($select);
+    }
+
+    /**
+     * @param array $productIds
+     * @return array
+     */
+    public function fetchImageData($productIds = [])
```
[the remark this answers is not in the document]
dawiddczaja: udpated

view/frontend/templates/menu/node_type/product.phtml
talalus: removed

view/adminhtml/web/vue/menu-type/product.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+    <div class="admin__field field">
+        <simple-field
+            :label="config.translation.productId"
+            id="snowmenu_node_product"
```
Igloczek: This should be dynamic, to avoids ID duplicates.

PR #62, opened 2018-07-11
[DEV-42285]

dawiddczaja:
[DEV-42285]

### DIFF
--- a/Api/Data/NodeInterface.php
+++ b/Api/Data/NodeInterface.php
@@ -3,5 +3,215 @@ namespace Snowdog\Menu\Api\Data;
 
 interface NodeInterface
 {
+    /**
+     * Constants for field names.
+     */
+    const NODE_ID = 'node_id';
+    const MENU_ID = 'menu_id';
+    const TYPE = 'type';
+    const CONTENT = 'content';
+    const CLASSES = 'classes';
+    const PARENT_ID = 'parent_id';
+    const POSITION = 'position';
+    const LEVEL = 'level';
+    const TITLE = 'title';
+    const TARGET = 'target';
+    const CREATION_TIME = 'creation_time';
+    const UPDATE_TIME = 'update_time';
+    const IS_ACTIVE = 'is_active';
 
+    /**
+     * Get node id
+     *
+     * @return int
+     */
+    public function getNodeId();
+
+    /**
+     * Set node id
+     *
+     * @param int $nodeId
+     * @return $this
+     */
+    public function setNodeId($nodeId);
+
+    /**
+     * Get menu id
+     *
+     * @return int
+     */
+    public function getMenuId();
+
+    /**
+     * Set menu id
+     *
+     * @param int $menuId
+     * @return $this
+     */
+    public function setMenuId($menuId);
+
+    /**
+     * Get type
+     *
+     * @return string
+     */
+    public function getType();
+
+    /**
+     * Set type
+     *
+     * @param string $type
+     * @return $this
+     */
+    public function setType($type);
+
+    /**
+     * Get content
+     *
+     * @return string
+     */
+    public function getContent();
+
+    /**
+     * Set content
+     *
+     * @param string $content
+     * @return $this
+     */
+    public function setContent($content);
+
+    /**
+     * Get classes
+     *
+     * @return string
+     */
+    public function getClasses();
+
+    /**
+     * Set classes
+     *
+     * @param string $classes
+     * @return $this
+     */
+    public function setClasses($classes);
+
+    /**
+     * Get parent id
+     *
+     * @return int
+     */
+    public function getParentId();
+
+    /**
+     * Set parent id
+     *
+     * @param int $parentId
+     * @return $this
+     */
+    public function setParentId($parentId);
+
+    /**
+     * Get position
+     *
+     * @return int
+     */
+    public function getPosition();
+
+    /**
+     * Set position
+     *
+     * @param int $position
+     * @return $this
+     */
+    public function setPosition($position);
+
+    /**
+     * Get level
+     *
+     * @return int
+     */
+    public function getLevel();
+
+    /**
+     * Set level
+     *
+     * @param int $level
+     * @return $this
+     */
+    public function setLevel($level);
+
+    /**
+     * Get title
+     *
+     * @return string
+     */
+    public function getTitle();
+
+    /**
+     * Set title
+     *
+     * @param string $title
+     * @return $this
+     */
+    public function setTitle($title);
+
+    /**
+     * Get target
+     *
+     * @return int
+     */
+    public function getTarget();
+
+    /**
+     * Set target
+     *
+     * @param int $target
+     * @return $this
+     */
+    public function setTarget($target);
+
+    /**
+     * Get creation time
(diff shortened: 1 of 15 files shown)


Api/MenuRepositoryInterface.php
```diff
@@ -6,15 +6,43 @@
...
 interface MenuRepositoryInterface 
 {
+    /**
+     * @param \Snowdog\Menu\Api\Data\MenuInterface $page
+     * @return \Snowdog\Menu\Api\Data\MenuInterface
+     */
     public function save(MenuInterface $page);
 
+    /**
+     * @param $id
```
jochhop: Please define type here

Model/Menu.php
```diff
@@ -45,4 +45,151 @@ public function saveStores(array $stores)
         }
         $connection->commit();
     }
+
+    /**
+     * Gets menu id.
+     *
+     * @return int|null
+     */
```
jochhop: In modules, I think it's better to use:

```php
/**
 * @inheritdoc
 */
```

so in case something changes in the interface we don't need to modify here too

Model/Menu/Node.php
```diff
@@ -14,8 +14,284 @@ protected function _construct()
...
     public function getIdentities()
     {
         return [self::CACHE_TAG . '_' . $this->getId()];
     }
+
+    /**
+     * Get node id
+     *
+     * @return int
+     */
```
jochhop: Same here, I think `@inheritdoc` is a better approach

Api/NodeRepositoryInterface.php
```diff
@@ -6,15 +6,42 @@
...
     public function deleteById($id);
 
+    /**
+     * Return node by menu id
+     *
+     * @api
+     * @param int $menuId
+     * @return \Snowdog\Menu\Api\Data\NodeInterface[]
+     */
     public function getByMenu($menuId);
```
jochhop: Should't this method be renamed to `getByMenuId`?

Api/MenuRepositoryInterface.php
dawiddczaja: updated

Api/NodeRepositoryInterface.php
dawiddczaja: updated

Model/Menu.php
dawiddczaja: updated

Model/Menu/Node.php
dawiddczaja: updated

Koc:
Thank you so much for this PR! :tada: 
I'm waiting for rest API for integration with headless Magento.

Api/Data/NodeInterface.php
```diff
@@ -3,5 +3,215 @@
...
+     */
+    public function getCreationTime();
+
+    /**
+     * Set creation time
+     *
+     * @param string $creationTime
+     * @return $this
+     */
+    public function setCreationTime($creationTime);
```
Koc: Does Magento cann't properly handle real `DateTime`? Plain strings looks a little bit wierd. Even more - we don't know what format to use here,

Api/NodeRepositoryInterface.php
```diff
@@ -6,15 +6,42 @@
...
 
-    public function getByMenu($menuId);
+    /**
+     * Return node by menu id
+     *
+     * @api
+     * @param int $menuId
+     * @return \Snowdog\Menu\Api\Data\NodeInterface[]
+     */
+    public function getByMenuId($menuId);
```
Koc: This change should be documented in changelog because BC breaks. Also why not use real typehinting for `int`?

DariuszMaciejewski: Yep, we shouldn't rename this method without a good reason.

We don't want to use type hints yet, because we have a few Magento 2.1 projects and I know there are issues with that (for example https://github.com/magento/magento2/issues/7620).

Api/Data/SearchItemResultsInterface.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Snowdog\Menu\Api\Data;
+
+use Magento\Framework\Api\SearchResultsInterface;
+
+/**
+ * Search results interface.
+ *
+ * @api
+ */
+interface SearchItemResultsInterface extends SearchResultsInterface
```
DariuszMaciejewski: Shouldn't be `Snowdog\Menu\Api\Data\MenuSearchResultsInterface` ? Remember about **di.xml**

Api/NodeRepositoryInterface.php
```diff
@@ -6,15 +6,42 @@
...
 
+    /**
+     * @param int $id
+     * @return \Snowdog\Menu\Model\Menu\Node
+     */
     public function getById($id);
 
+    /**
+     * @param \Magento\Framework\Api\SearchCriteriaInterface $criteria
+     * @return \Magento\Framework\Api\SearchResultsInterface
```
DariuszMaciejewski: Shouldn't we create a custom interface for that (like you did for menus) ? I mean `Snowdog\Menu\Api\Data\NodeSearchResultsInterface`.

Api/NodeRepositoryInterface.php
dawiddczaja: function name reverted

Api/NodeRepositoryInterface.php
dawiddczaja: I added `NodeSearchResultsInterface/MenuSearchResultsInterface` because I use `MenuRepositoryInterface::getList` function to return data by api, without `NodeSearchResultsInterface/MenuSearchResultsInterface` api return empty arrays without data. In this task I had to do two api endpoint this function `NodeRepositoryInterface::getList` is not related with this task but I can add it

Api/Data/NodeInterface.php
dawiddczaja: Magento can not handle `DATETIME` so I left the type and added the format

Api/MenuRepositoryInterface.php
```diff
@@ -6,15 +6,43 @@
...
+     * @return \Snowdog\Menu\Model\Menu
+     */
     public function getById($id);
 
+    /**
+     * Returns menus list
+     *
+     * @api
+     * @param \Magento\Framework\Api\SearchCriteriaInterface $criteria
+     * @return \Snowdog\Menu\Api\Data\MenuSearchResultsInterface|string
```
DariuszMaciejewski: I think we should update the return object (type) in our implementation of this interface too: https://github.com/SnowdogApps/magento2-menu/blob/master/Model/MenuRepository.php#L66

Api/NodeRepositoryInterface.php
DariuszMaciejewski: okay, maybe some other time

Api/MenuRepositoryInterface.php
dawiddczaja: Updated

Koc:
Why not merged? This is very needed feature

dawiddczaja:
@Koc please check https://github.com/SnowdogApps/magento2-menu/commits/develop

Issue #63, opened 2018-08-03
Menu Node Name field lack a translation mechanism


gtlt:
In multi language storeview setup, we need to define a menu for each storeview in order to translate Node Name. Only for CmsBlock type it is possible to use the "trans" directive inside the block it.
Do you think it would be possible to add a translation mechanism for Node Name to avoid maintaining n menu across all storeviews ?

gtlt:
Well the change is quite simple, I made a pr #64 for review.
Thank you

Igloczek:
@talalus could you check this out? IMO the category names should be loaded from M categories, so if everything is set well in here, it should be reflected in the menu too, without the need of using translations as proposed in #64.

talalus:
1. I think it's better to give the freedom to define a custom node name for the category node because in some cases the menu node name might be different from category name, but I think it's a good Idea to auto fill the node name with category name after selecting a category.
2. Will check with backend Dev how much is complicated to make a menu item has node name per store view

Igloczek:
@gtlt I think the simples solution for you is to overwrite catalog node template on theme level and replace this line

https://github.com/SnowdogApps/magento2-menu/blob/1e538845941bbaeaa83a77c0b10aab493ccedfd8/view/frontend/templates/menu/node_type/category.phtml#L28

with
```php
<?= __($title) ?>
```

or
```php
<?= __('unique-menu-name' . $title) ?>
```
to distinguish menu translations form global one.

Apart from that, we are planning to add "Duplicate menu" button, to simplify the process of creating multiple similar menus.

gtlt:
thank you, will use that

PR #64, opened 2018-08-03
Translate Category Node Name (Title)

gtlt:


Igloczek:
It's quite a breaking change, can break any existing menu in a very unpredictable way, just by using translations not intended to be used by the menu.

Discussion about possible ways to solve this problem is in #63

Issue #44, opened 2018-02-07
Target Node Blank Not Saving


realchriswells:
With Snowdog Menu 2.4.0 there is an option for Target Node Blank, which I assume adds `target="_blank"` to a link in the menu. I have checked this option, but it seems to not save anything as when the page refreshes, the option is blank again and after a `bin/magento setup:upgrade` the database doesn't seem to have anywhere to hold the field.

Igloczek:
Internal ticket DEV-38867

realchriswells:
I managed to fix this issue directly in the database.

It seems that it is the field that does not get created, and it seems to be the wrong type - of course I'm just speculating at this point. I'm basing it on the fact that in the `UpgradeSchema.php` file, the `addTargetAttribute()` asks to create a field of type `Table::TYPE_TEXT`, a length of `10` and a `default` value. When I tried to manually create it in the database, I couldn't add the `length` and `default` values to the field. Changing the type to `VARCHAR` resolved this.

talalus:
@dawiddczaja should be fixed in the admin panel, already merged with develop

PR #66, opened 2018-08-23
Fix true/false values for target checkbox

haelbichalex:
#44: Checking the target _blank option didn't save, because of the old true/false values used in the .vue file.


haelbichalex:
Can you please tag a new release with this fix? This is open since August 👍

talalus:
you can use 2.8.0 now

PR #77, opened 2019-02-12
Fix for missing url when a cms page has no url rewrites

### LINKED ISSUE #76: Error in cms page renderer when there are no url_rewrites for a cms page
CMS Page Node renderer doesn't check if a URL for a particular page exists in `pageUrls` array:
https://github.com/SnowdogApps/magento2-menu/blob/1e538845941bbaeaa83a77c0b10aab493ccedfd8/Block/NodeType/CmsPage.php#L151-L168

Apparently, such an event can happen because `pageUrls` array is hydrated from `url_rewrite` table data When a cms page doesn't have any URL rewrites for that store, then its URL is not included in the `pageUrls` array.

https://github.com/SnowdogApps/magento2-menu/blob/1e538845941bbaeaa83a77c0b10aab493ccedfd8/Model/ResourceModel/NodeType/CmsPage.php#L58-L71

In result we're getting:

```
1 exception(s):
Exception #0 (Exception): Notice: Undefined offset: 3 in /var/www/magento/vendor/snowdog/module-menu/Block/NodeType/CmsPage.php on line 158
```

A solution would be to back fetched data up with `identifier` from `cms_page` joined with `cms_page_store`

lukeromanowicz:
Resolves #76

### DIFF
--- a/Model/ResourceModel/NodeType/CmsPage.php
+++ b/Model/ResourceModel/NodeType/CmsPage.php
@@ -15,6 +15,7 @@ use Magento\Cms\Api\PageRepositoryInterface;
 use Magento\Framework\Api\SearchCriteriaBuilder;
 use Magento\Framework\App\ResourceConnection;
 use Magento\Framework\EntityManager\MetadataPool;
+use Magento\Framework\Exception\LocalizedException;
 use Magento\Store\Model\Store;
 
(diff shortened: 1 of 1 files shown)


lukeromanowicz:
@kiber absolutely. All requested changes have been made.

PR #86, opened 2019-05-06
#53114 Prepending store url in Custom Url nodes should be optional

Ramiz4ik:


jochhop:
IMO the changes don't resolve the issue, still there is store URL prefixed to the url filled in node. So I would not change the template itself, and add just a comparison in `getHtml()` method like:

```php
$nodeContent  = $node->getContent();
if (!$this->isExternalUrl($nodeContent)) {
    $url = $this->_storeManager->getStore()->getBaseUrl() . $nodeContent;
} else {
    $url = $nodeContent;
}
```

Block/NodeType/CustomUrl.php
```diff
@@ -81,14 +81,31 @@ public function getHtml($nodeId, $level)
     {
         $classes = $level == 0 ? 'level-top' : '';
         $node = $this->nodes[$nodeId];
-        $url = $this->_storeManager->getStore()->getBaseUrl() . $node->getContent();
+        $nodeContent  = $node->getContent();
         $title = $node->getTitle();
+        $target = $this->getTarget() == 1 ? 'target="_blank"' : '';
```
dariuszmaciejewski: The comparison is not needed.

Block/NodeType/CustomUrl.php
```diff
@@ -81,14 +81,31 @@ public function getHtml($nodeId, $level)
...
+<a href="$url" class="$classes" $target role="menuitem"><span>$title</span></a>
 HTML;
     }
 
+    /**
+     * @param string $url
+     *
+     * @return bool
+     */
+    private function isExternalUrl(string $url = '')
```
dariuszmaciejewski: The default value `$url` is not needed.  BTW are you sure the `$node->getContent()` method always return `string`? I ask because you use type hint and it will cause fatal error for `null` as the parameter.

Block/NodeType/CustomUrl.php
Ramiz4ik: Done

Block/NodeType/CustomUrl.php
Ramiz4ik: Done

PR #87, opened 2019-05-07
#53115 Add getCategory method

Ramiz4ik:


Block/NodeType/Category.php
```diff
@@ -150,6 +159,24 @@ public function getCategoryUrl($nodeId, $storeId = null)
...
+    {
+        if (!isset($this->nodes[$nodeId])) {
+            throw new \InvalidArgumentException('Invalid node identifier specified');
+        }
+
+        $storeId = $this->_storeManager->getStore()->getId();
+        $node = $this->nodes[$nodeId];
+        $categoryId = (int) $node->getContent();
+
+        return $this->categoryRepository->get($categoryId, $storeId);
```
dariuszmaciejewski: Do you load a category in a loop here (which renders a single node)? I think isn't the best idea. Can we load all needed categories as a collection before rendering our menu?

Ramiz4ik: Done

Model/NodeType/Category.php
```diff
@@ -97,16 +106,26 @@ public function fetchData(array $nodes, $storeId)
...
         $this->profiler->stop(__METHOD__);
 
-        return [$localNodes, $categoryUrls];
+        return [$localNodes, $categoryUrls, $categories];
+    }
+
+    public function getCategory($categoryId)
+    {
+        $category = $this->categoryFactory->create();
+        $category->load($categoryId);
```
dariuszmaciejewski: As I said before - use a category collection to load all needed categories from DB at once.

Model/NodeType/Category.php
```diff
@@ -97,16 +106,26 @@ public function fetchData(array $nodes, $storeId)
 
         $localNodes = [];
         $categoryIds = [];
+        $categories = [];
 
         foreach ($nodes as $node) {
             $localNodes[$node->getId()] = $node;
-            $categoryIds[] = (int)$node->getContent();
+            $categoryId = (int)$node->getContent();
+            $categoryIds[] = $categoryId;
```
dariuszmaciejewski: You can use `array_keys($categories)` instead later.

Model/NodeType/Category.php
```diff
@@ -104,9 +113,33 @@ public function fetchData(array $nodes, $storeId)
...
+        $categories = $this->getCategories($storeId, $categoryIds);
 
         $this->profiler->stop(__METHOD__);
 
-        return [$localNodes, $categoryUrls];
+        return [$localNodes, $categoryUrls, $categories];
+    }
+
+    /**
+     * @param int $storeId
```
jochhop: From what I see in `setStore` method definition, the method accepts all these types:

```php
@param int|string|\Magento\Store\Model\Store $store
```

So please update parameter name to be `$store` instead and add the accepted types to the annotations (not only `int`)

Ramiz4ik: Done

PR #89, opened 2019-05-10
WIP: Fixes #83: It is now possible to access category entity within category node template

### LINKED ISSUE #83: Provide access to entity associated with menu node
For the second time I ran into a case where I would need to access category data in the template (`menu/node_type/category.phtml`).

I can get for example category URL with `getCategoryUrl()`, but I'm not able to get the category object itself.

Would it be OK to expose the category e.g. with `\Snowdog\Menu\Block\NodeType\Category::getCategory()`?

Or perhaps it could even be called `getEntity()` so the same method could be implemented also for Product, CmsPage and CmsBlock node types?

juho-jaakkola:
Implements a new method `\Snowdog\Menu\Block\NodeType\Category::getCategory()` that
can be used to access the category entity within the block template.

Example usage:

```
/** @var \Snowdog\Menu\Block\NodeType\Category $block */
$imageSrc = $block->getCategory($nodeid)->getImageUrl();
```

### DIFF
--- a/Api/Data/NodeInterface.php
+++ b/Api/Data/NodeInterface.php
@@ -10,6 +10,7 @@ interface NodeInterface
     const MENU_ID = 'menu_id';
     const TYPE = 'type';
     const CONTENT = 'content';
+    const ENTITY = 'entity';
     const CLASSES = 'classes';
     const PARENT_ID = 'parent_id';
     const POSITION = 'position';
--- a/Block/NodeType/Category.php
+++ b/Block/NodeType/Category.php
@@ -2,6 +2,7 @@
 
 namespace Snowdog\Menu\Block\NodeType;
 
+use Magento\Catalog\Model\CategoryRepository;
 use Magento\Framework\View\Element\Template\Context;
 use Magento\Framework\Registry;
 use Snowdog\Menu\Model\TemplateResolver;
@@ -27,6 +28,10 @@ class Category extends AbstractNode
      */
     protected $categoryUrls;
     /**
+     * @var array
+     */
+    protected $categories;
+    /**
      * @var Registry
      */
     private $coreRegistry;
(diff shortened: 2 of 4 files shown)


juho-jaakkola:
I am not completely sure if this is the best approach, so therefore its still marked as WIP.

Api/Data/NodeInterface.php
```diff
@@ -10,6 +10,7 @@ interface NodeInterface
     const MENU_ID = 'menu_id';
     const TYPE = 'type';
     const CONTENT = 'content';
+    const ENTITY = 'entity';
```
juho-jaakkola: This is currently not needed. It got left behind when experimented on adding the `getCategory()` method into the Node object instead of the Block object.

However, now that I noticed it, I think it *would* in fact be much better from the API user's point of view if the method was implemented into the Node object.

The reason I ended up adding the method to the block is that also the [getCategoryUrl()](https://github.com/SnowdogApps/magento2-menu/blob/master/Block/NodeType/Category.php#L134-L151) has been implemented in it.

Why is this? Why not just make it possible to call `$node->getUrl()`? And respectively `$node->getCategory()`?

Block/NodeType/Category.php
```diff
@@ -102,7 +110,7 @@ public function fetchData(array $nodes)
     {
         $storeId = $this->_storeManager->getStore()->getId();
 
-        list($this->nodes, $this->categoryUrls) = $this->_categoryModel->fetchData($nodes, $storeId);
+        list($this->nodes, $this->categoryUrls, $this->categories) = $this->_categoryModel->fetchData($nodes, $storeId);
```
juho-jaakkola: The `$this->categoryUrls` property becomes obsolete if the rest of the changes are approved. I left it in until I've gotten some feedback on whether my approach is acceptable in the first place.

juho-jaakkola:
Already implemented in release `2.10.0`, so closing the merge.

See https://github.com/SnowdogApps/magento2-menu/compare/2.9.0...2.10.0

Issue #85, opened 2019-04-26
Status for node "enabled/disabled"


Januszpl:
Don't see this option now and I think it should be as then user doesn't need to remove something and later add as it can just disable node

andrzejborkowski:
the feature has been released within [2.13.0 tag](https://github.com/SnowdogApps/magento2-menu/tree/2.13.0)

PR #103, opened 2020-03-30
#85 add is_active into menu data

kiber:
DEV-65561


view/adminhtml/web/vue/field-type/checkbox.vue
```diff
@@ -0,0 +1,40 @@
...
+                fieldId: '',
+                checkboxValue: this.value
+            }
+        },
+        mounted: function() {
+            this.fieldId = 'snowmenu_' + this.id + '_' + this._uid;
+        },
+        watch: {
+          checkboxValue: function(newValue, oldValue) {
+            this.item.is_active = newValue ? '1' : '0'
```
Igloczek, line 35: We shouldn't mutate the prop directly. If it's just an UI thing move it to computed.

Januszpl: Yes. I agree. But we would first to refactor few things related to saving whole menu data. and I think it should be done in a new ticket

Model/Menu/NodeRepository.php
```diff
@@ -29,6 +38,11 @@ public function __construct(
         $this->searchResultsFactory = $searchResultsFactory;
     }
 
+    /**
+     * @param NodeInterface $object
+     * @return NodeInterface
+     * @throws CouldNotSaveException
+     */
```
jochhop, line 45: I would use `@inheritdoc` instead (I see it's used already in Menu.php model class, so let's keep the same format)

kiber: Added changes for all methods

mcjwsk (approved):

mcjwsk (approved):

Issue #84, opened 2019-04-26
Be able to set node template per "node" and "submenu" template


Januszpl:
It would be great if we could set specific template for every node item and its submenu template if it has children

```
  <li class="list__item <?= $itemClasses; ?>">
            <?= $block->renderMenuNode($node); ?>
            <?= $block->renderSubmenu($children, $node, $childrenLevel); ?>
        </li>
```

I think we should pass extra argument $template,
in `renderMenuNode()`   as second argument (be default it would use default template depending on the node type).
in `renderSubmenu()` as fourth argument  (by default it would use default template sub_menu.phtml)

The best if we could select or set the path or name of template in admin panel in node setting. if template not found then use default.

**Why this feature?**

 sometimes menu structure and styling is different for different section (submenu). Now sometimes it requires adding extra logic in templates or/and classes in admin panel or code in css. Having possibility to have custom templates for nodes and submenu we can create more clean templates with specific classes and achieve really different look and structure if required.

It would be useful if we have one menu and structure like that:
Main-menu
 --- Section A
     (submenu template A:  `menu/sub_menu_section-a.phtml`)
     --- item A.1 (node custom_url, custom template: `menu/node_type/custom_url-section-a.phtml`
    --- itd . A.2 (node custom_url, default template:` menu/node_type/custom_url.phtml`
           ---      (submenu default template: `menu/sub_menu.phtml`

--- Section B
     (submenu template B: `menu/sub_menu_section-b.phtml`)
      --- item B.1 (node X, template: default or set via admin panel)

PR #109, opened 2020-07-20
Move submenu template path to protected variable

j-stam:
Recently i had to overwrite the constructor to change the submenu template and add a couple of private functions. This fix allows changing the submenu template without touching the constructor


snowremi:
Hi! I created an internal ticket DEV-69228 for this, and we'll work on this shortly

Januszpl:
There is also one ticket where we are working on adding functionality to admin panel so user can change node/submenu templates if there are any custom ".phtml" templates added in themes. 
However this change I think is still valid but backend developer should check this change with #84

PR #113, opened 2020-07-24
Feature: Extract responsibility of mapping request to menu model

lbajsarowicz:
With the current approach, when you want to add an extra attribute to Menu, you cannot achieve that without inheritance of `Save` controller, you cannot achieve that.

I have introduced the `Hydrator` service, that enables you to map Request elements to Menu model.
This way you can apply `after` plugin that maps extra fields to the menu.

This is the 1st step of upcoming changes, as I'd love to introduce such hydrator also for Nodes.

### DIFF
--- a/Controller/Adminhtml/Menu/Index.php
+++ b/Controller/Adminhtml/Menu/Index.php
@@ -4,14 +4,15 @@ declare(strict_types=1);
 namespace Snowdog\Menu\Controller\Adminhtml\Menu;
 
 use Magento\Backend\App\Action;
-use Magento\Backend\App\Action\Context;
 use Magento\Framework\Controller\ResultFactory;
-use Magento\Framework\View\Result\PageFactory;
 
 class Index extends Action
 {
     public const ADMIN_RESOURCE = 'Snowdog_Menu::menus';
 
+    /**
+     * @inheritDoc
+     */
     public function execute()
     {
         return $this->resultFactory->create(ResultFactory::TYPE_PAGE);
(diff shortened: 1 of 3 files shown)


snowremi:
Internal ticket DEV-69365

TM18:
@lbajsarowicz was this ticket closed on purpose?

lbajsarowicz:
A month passed by without any feedback.
I'm going to fork and develop the MegaMenu on my own.

lbajsarowicz:
giphy

Don't make it hang too long.

Controller/Adminhtml/Menu/Index.php
```diff
@@ -1,29 +1,19 @@
 <?php
+declare(strict_types=1);
+
 namespace Snowdog\Menu\Controller\Adminhtml\Menu;
 
 use Magento\Backend\App\Action;
 use Magento\Backend\App\Action\Context;
+use Magento\Framework\Controller\ResultFactory;
 use Magento\Framework\View\Result\PageFactory;
```
TM18, line 9: `Context` and `PageFactory` are not used anymore so I think they can be safely removed from here.

TM18 (changes requested):

Controller/Adminhtml/Menu/Index.php
lbajsarowicz: Yes, you are right. I'm going to adjust that later afternoon.

TM18 (approved):

PR #121, opened 2020-08-25
Feature: Validate only changed files, display errors inline

lbajsarowicz:
I modified some PHP file introducing some issue that is reported by PHPCS.

Also:
1. Checkout only `depth = 1` to reduce the cost of pulling recent changes
2. Migrated your Composer configuration
3. Introduced Composer cache
4. Executing only on modified files

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -1,24 +1,40 @@
-name: Backend
-on: [push, pull_request_target]
+---
+name: Backend Coding Standard
+on: [ push, pull_request ]
 jobs:
   backend-lint:
     runs-on: ubuntu-latest
+    name: MCS Check
+    container:
+      image: docker.io/davidalger/php:7.4
+
     steps:
-    - name: Checkout
-      uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v1
+        with:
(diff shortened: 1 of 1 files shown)


lbajsarowicz:
I couldn't force Github to run my `backend.yml` file from my own branch (it still uses `pull_request_target` as a source of tests).

Igloczek:
@lbajsarowicz do you know why it fails without throwing any error?
https://github.com/SnowdogApps/magento2-menu/runs/1030198800?check_suite_focus=true

lbajsarowicz:
@Igloczek I'm going to submit some failing file to check this on real changes. Then it should work, as it failed due to no files provided.

Igloczek:
It was working only for `pull_request` events, but not `push` due to `only_changed_files` option (on push there is no way to build a list of changed files, because is probably compare branches).

lbajsarowicz:
Actually it's written the way that it checks what the Push belongs to and fetches the changed files for all the PR:

https://github.com/Mediotype/phpcs-action/blob/980dd952ece08ad1a1baca4447df4a50ca4d7cd6/entrypoint.sh#L8-L13

I've been using that successfully on other projects.

Igloczek:
Weird :( It stops silently failing after switching that off.

PR #123, opened 2020-08-26
[70934] Commit some ugly file to test CS

lbajsarowicz:


Igloczek:
Looks like we need to back to `pull_request_target` events, because otherwise, the forks do not have access to secrets from this repository.

lbajsarowicz:
@Igloczek I had no idea about this limitation, but it makes sense!

lbajsarowicz:
@Igloczek I have no idea why it still does not work as expected, but I'll check that afternoon.
Sorry to bother you, but I believe that it brings huge value when developer knows what exactly he broke.

Igloczek:
Sure thing. I'll bring back also the previous solution, so temporarily we will have double checks, before the new one will be working as expected.

lbajsarowicz:
Sure, thanks!

Igloczek:
BTW. Aree you 100% sure that this code shouldn't pass the CS checks? The old solution is also not reporting any problems 🤔

lbajsarowicz:
It should fail due to multiple empty lines and long line close to the end of file.

emastyle:
We are closing this as it seems related to action/checkout@v2 (we are using in our workflow) issue that should be fixed in the next release.

PR #126, opened 2020-09-01
Added delete and massDelete methods to menu listing

kiber:
DEV-69083

### DIFF
--- a/Controller/Adminhtml/Menu/Delete.php
+++ b/Controller/Adminhtml/Menu/Delete.php
@@ -90,7 +90,7 @@ class Delete extends Action
 
         try {
             $menu = $this->menuRepository->getById($menuId);
-            $this->menuRepository->deleteById($menuId);
+            $this->menuRepository->delete($menu);
 
             $filterBuilder = $this->filterBuilderFactory->create();
             $filter = $filterBuilder->setField('menu_id')
--- a/Controller/Adminhtml/Menu/Delete.php
+++ b/Controller/Adminhtml/Menu/Delete.php
@@ -10,6 +10,7 @@ use Magento\Framework\Api\Search\FilterGroupBuilderFactory;
 use Magento\Framework\Api\SearchCriteriaBuilderFactory;
 use Magento\Framework\App\ResponseInterface;
 use Magento\Framework\Exception\CouldNotDeleteException;
+use Magento\Framework\Exception\NoSuchEntityException;
 use Snowdog\Menu\Api\MenuRepositoryInterface;
 use Snowdog\Menu\Api\NodeRepositoryInterface;
 use Magento\Framework\Controller\ResultInterface;
@@ -109,8 +110,10 @@ class Delete extends Action
                 $this->nodeRepository->delete($node);
             }
             $this->messageManager->addSuccessMessage(__('Menu %1 and it\'s nodes removed', $menu->getTitle()));
-        } catch (CouldNotDeleteException $e) {
-            $this->messageManager->addErrorMessage($e->getMessage());
+        } catch (NoSuchEntityException $exception) {
+            $this->messageManager->addErrorMessage($exception->getMessage());
+        } catch (CouldNotDeleteException $exception) {
+            $this->messageManager->addErrorMessage($exception->getMessage());
         }
 
         $redirect = $this->resultRedirectFactory->create();
--- a/Controller/Adminhtml/Menu/Delete.php
+++ b/Controller/Adminhtml/Menu/Delete.php
@@ -17,6 +17,8 @@ use Magento\Framework\Exception\NotFoundException;
 
 /**
  * Class Delete
+ *
+ * This action deletes single menu
  */
 class Delete extends Action
 {
--- a/Controller/Adminhtml/Menu/Delete.php
+++ b/Controller/Adminhtml/Menu/Delete.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Snowdog\Menu\Controller\Adminhtml\Menu;
 
 use Magento\Backend\App\Action;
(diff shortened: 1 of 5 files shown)


Controller/Adminhtml/Menu/MassDelete.php
```diff
@@ -0,0 +1,78 @@
...
+     * @return Redirect
+     * @throws LocalizedException|Exception
+     */
+    public function execute()
+    {
+        $collection = $this->filter->getCollection($this->collectionFactory->create());
+        $collectionSize = $collection->getSize();
+
+        foreach ($collection as $menu) {
+            $menu->delete();
```
jochhop, line 68: Please use `\Snowdog\Menu\Api\MenuRepositoryInterface`'s `delete` as `delete` from `AbstractModel` is deprecated.

Controller/Adminhtml/Menu/MassDelete.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Execute action
+     *
+     * @return Redirect
+     * @throws LocalizedException|Exception
+     */
+    public function execute()
+    {
+        $collection = $this->filter->getCollection($this->collectionFactory->create());
```
jochhop, line 64: I'd move this into a private method, so we can just call it eg. `$this->getMenuCollection()`

Controller/Adminhtml/Menu/MassDelete.php
```diff
@@ -0,0 +1,78 @@
...
+        $collectionSize = $collection->getSize();
```
jochhop, line 65: I'd move this after the `foreach`, this way if something goes wrong during menu removal this variable does not initialize

Controller/Adminhtml/Menu/MassDelete.php
kiber: Done

Controller/Adminhtml/Menu/MassDelete.php
kiber: Done

Controller/Adminhtml/Menu/MassDelete.php
kiber: I have tested your suggestion but it returns 0 after `foreach` loop.

jochhop: Yeah... my bad, it's removing it from db and from collection, sorry 🤦

Controller/Adminhtml/Menu/Delete.php
```diff
@@ -53,19 +81,22 @@ public function __construct(
...
     public function execute()
     {
-        $id = $this->getRequest()->getParam('id');
+        $menuId = (int) $this->getRequest()->getParam('id');
 
         try {
-            $menu = $this->menuRepository->getById($id);
-            $this->menuRepository->deleteById($id);
+            $menu = $this->menuRepository->getById($menuId);
+            $this->menuRepository->deleteById($menuId);
```
jochhop, line 93: I think we could use `$this->menuRepository->delete($menu)`, this way we avoid loading a menu twice

Ui/Component/Listing/Column/MenuList/PageActions.php
```diff
@@ -1,26 +1,93 @@
...
-                $name = $this->getData("name");
-                $id = "X";
-                if (isset($item["menu_id"])) {
-                    $id = $item["menu_id"];
+        if (isset($dataSource['data']['items'])) {
+            foreach ($dataSource['data']['items'] as & $item) {
+                if (isset($item['menu_id'])) {
+                    $name = $this->getData('name');
+                    $item[$name] = [
+                        'edit' => [
```
jochhop, line 64: This method seems pretty long, what do you think if we extract `edit` and `delete` buttons generation into 2 different private methods so this part will look like:

```php
// ...
$item[$name] = [
    'edit' => $this->getEditButton($item['menu_id']),
    'delete' => $this->getDeleteButton($item['menu_id']),    
]
// ...
```

Controller/Adminhtml/Menu/MassDelete.php
```diff
@@ -0,0 +1,97 @@
...
+     * Authorization level of a basic admin session
+     *
+     * @see _isAllowed()
+     */
+    const ADMIN_RESOURCE = 'Snowdog_Menu::menus';
+
+    /**
+     * @var Filter
+     */
+    protected $filter;
```
jochhop, line 36: Is there any reason why these properties are set as `protected`, it looks like they are not going to be overriden/used in a child class, I think we could switch them to `private` instead.

kiber: There are other properties that are `protected`, that is why I left it also as `protected`. Now I changed it to `private`

Ui/Component/Listing/Column/MenuList/PageActions.php
kiber: Done

Controller/Adminhtml/Menu/Delete.php
kiber: Done

jochhop (dismissed):

Controller/Adminhtml/Menu/Delete.php
```diff
@@ -53,19 +81,22 @@ public function __construct(
...
      */
     public function execute()
     {
-        $id = $this->getRequest()->getParam('id');
+        $menuId = (int) $this->getRequest()->getParam('id');
 
         try {
-            $menu = $this->menuRepository->getById($id);
-            $this->menuRepository->deleteById($id);
+            $menu = $this->menuRepository->getById($menuId);
```
TM18, line 92: `->getById($menuId)` throws a `NoSuchEntityException`, I wonder if we should catch it as well, same as with `CouldNotDeleteException`. I see both extend the `LocalizedException`, maybe catch this one only, what do you think?

Controller/Adminhtml/Menu/MassDelete.php
```diff
@@ -0,0 +1,97 @@
...
+     * @return Redirect
+     * @throws LocalizedException|Exception
+     */
+    public function execute()
+    {
+        $collection = $this->getMenuCollection();
+        $collectionSize = $collection->getSize();
+
+        foreach ($collection as $menu) {
+            $this->menuRepository->delete($menu);
```
TM18, line 78: Shouldn't we catch `CouldNotDeleteException` here as well?

kiber: Done

Controller/Adminhtml/Menu/Delete.php
kiber: Added another line to catch `NoSuchEntityException` exception. I think this way more clear than using one `LocalizedException` exception.

jochhop (approved):

PR #127, opened 2020-09-07
Fixed v-if check on option.store array

reense:
We've had this problem with some invalid pages. option.store is always an array, however, it might be empty.
With this merge request we explicitly check if there are any elements in the array.

### DIFF
--- a/view/adminhtml/web/vue/field-type/autocomplete.vue
+++ b/view/adminhtml/web/vue/field-type/autocomplete.vue
@@ -14,7 +14,7 @@
                 <template v-slot:option="option">
                     {{ option.label }}
 
-                    <template v-if="option.store">
+                    <template v-if="option.store.length > 0">
                         <span class="vs__dropdown-option__details">
                             {{ option.store.join(', ') }}
                         </span>


Januszpl (approved):

Januszpl:
@reense Thank you for your contribution. Your changes have been merged.

Januszpl:
@reense I had to revert this change as it was breaking other use cases. However I will take look into this issue.

Januszpl:
Can you tell us more about 'some invalid pages.' ? How I could reproduce the issue? I think the only problem can be if somehow store is not set for cms pages but as far I've checked this field is required in magento2 admin panel.

reense:
@Januszpl That sucks.. I apologize. After an import some pages were missing entries in `cms_page_store`, so there was no store bound to those pages, so the "Stores" tab on those pages was empty. So when loading the menu builder, it crashes since `page.store` was null.

Januszpl:
@reense I think it should be fixed on db level with sql query as this is wrong import. However I agree that even tough menu shouldn't crash this way so I will check later how to handle this kind of situation or I will just improve your fix so it will handle all cases. Just didn't have time to look into and had to fix current develop.

PR #128, opened 2020-09-10
[70191] Invalidate page cache on menu changes

### LINKED ISSUE #115: Invalidate full page cache
After you edit the menu with caching enabled, it is necessary to refresh the "Page Cache" before the changes take effect on the frontend.

It would be useful if when the menu is edited the page cache is invalidated so that the admin knows that it required flushing:
https://www.atwix.com/development/managing-cache-programmatically/

akai-z:
Resolves #115

### DIFF
--- a/Controller/Adminhtml/Menu/Save.php
+++ b/Controller/Adminhtml/Menu/Save.php
@@ -95,7 +95,7 @@ class Save extends Action
     {
         $menu = $this->getCurrentMenu();
 
-        $menu->setIsActive(1);
+        $menu->setIsActive('1');
         $this->hydrator->mapRequest($menu, $this->getRequest());
         $menu = $this->menuRepository->save($menu);
 
@@ -194,9 +194,9 @@ class Save extends Action
 
                     $nodeObject->setMenuId($menu->getMenuId());
                     $nodeObject->setTitle($node['title']);
-                    $nodeObject->setIsActive((int) ($node['is_active'] ?? 0));
-                    $nodeObject->setLevel($level);
-                    $nodeObject->setPosition($position);
+                    $nodeObject->setIsActive($node['is_active'] ?? '0');
+                    $nodeObject->setLevel((string) $level);
+                    $nodeObject->setPosition((string) $position);
 
                     $this->nodeRepository->save($nodeObject);
 
--- a/Model/Menu.php
+++ b/Model/Menu.php
@@ -18,30 +18,6 @@ class Menu extends AbstractModel implements MenuInterface, IdentityInterface
 
     protected $_cacheTag = self::CACHE_TAG;
 
-    /**
-     * @var Menu\Cache
-     */
-    private $cache;
-
-    public function __construct(
-        Context $context,
-        Registry $registry,
-        Menu\Cache $cache,
-        AbstractResource $resource = null,
-        AbstractDb $resourceCollection = null,
-        array $data = []
-    ) {
-        $this->cache = $cache;
-
-        parent::__construct(
-            $context,
-            $registry,
-            $resource,
-            $resourceCollection,
-            $data
-        );
-    }
-
     protected function _construct()
     {
         $this->_init(\Snowdog\Menu\Model\ResourceModel\Menu::class);
--- a/Model/Menu.php
+++ b/Model/Menu.php
@@ -1,8 +1,12 @@
 <?php
 namespace Snowdog\Menu\Model;
 
+use Magento\Framework\Data\Collection\AbstractDb;
 use Magento\Framework\DataObject\IdentityInterface;
 use Magento\Framework\Model\AbstractModel;
+use Magento\Framework\Model\Context;
+use Magento\Framework\Model\ResourceModel\AbstractResource;
+use Magento\Framework\Registry;
 use Snowdog\Menu\Api\Data\MenuInterface;
 
 /**
@@ -13,6 +17,30 @@ class Menu extends AbstractModel implements MenuInterface, IdentityInterface
     const CACHE_TAG = 'snowdog_menu_menu';
 
     protected $_cacheTag = self::CACHE_TAG;
+
+    /**
+     * @var Menu\Cache
+     */
+    private $cache;
+
+    public function __construct(
+        Context $context,
+        Registry $registry,
+        Menu\Cache $cache,
+        AbstractResource $resource = null,
+        AbstractDb $resourceCollection = null,
+        array $data = []
+    ) {
+        $this->cache = $cache;
+
+        parent::__construct(
+            $context,
+            $registry,
+            $resource,
+            $resourceCollection,
+            $data
+        );
+    }
 
     protected function _construct()
     {
@@ -48,6 +76,7 @@ class Menu extends AbstractModel implements MenuInterface, IdentityInterface
             $connection->insert($table, ['menu_id' => $this->getId(), 'store_id' => $store]);
         }
         $connection->commit();
+        $this->cache->invalidatePageCache();
     }
 
     /**
--- a/Model/Menu.php
+++ b/Model/Menu.php
@@ -76,7 +76,6 @@ class Menu extends AbstractModel implements MenuInterface, IdentityInterface
             $connection->insert($table, ['menu_id' => $this->getId(), 'store_id' => $store]);
         }
         $connection->commit();
-        $this->cache->invalidatePageCache();
     }
 
     /**
--- a/Model/Menu.php
+++ b/Model/Menu.php
@@ -36,6 +36,10 @@ class Menu extends AbstractModel implements MenuInterface, IdentityInterface
 
     public function saveStores(array $stores)
     {
+        if ($stores == $this->getStores()) {
+            return;
+        }
+
         $connection = $this->getResource()->getConnection();
         $connection->beginTransaction();
         $table = $this->getResource()->getTable('snowmenu_store');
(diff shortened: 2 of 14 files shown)


Model/Menu.php
```diff
@@ -36,6 +64,10 @@ public function getStores()
 
     public function saveStores(array $stores)
     {
+        if ($stores == $this->getStores()) {
```
jochhop, line 67: Is this really necessary? I have seen a few lines below that all entries related to the current menu are removed from the table and then they're reinserted: https://github.com/SnowdogApps/magento2-menu/blob/develop/Model/Menu.php#L42

Model/Menu/NodeRepository.php
```diff
@@ -120,6 +130,7 @@ public function getList(SearchCriteriaInterface $criteria)
         $collection->setPageSize($criteria->getPageSize());
         $objects = [];
         foreach ($collection as $objectModel) {
+            $objectModel->setHasDataChanges(false);
```
jochhop, line 133: I have a question about this line, I have tested the module, and this line doesn't make any difference (at least during my test cases), is there any reason why was it added? maybe I am missing some case and this is useful to avoid issues?

Model/MenuRepository.php
```diff
@@ -133,6 +142,7 @@ public function getList(SearchCriteriaInterface $criteria)
         $collection->setPageSize($criteria->getPageSize());
         $objects = [];
         foreach ($collection as $objectModel) {
+            $objectModel->setHasDataChanges(false);
```
jochhop, line 145: I have a question about this line, I have tested the module, and this line doesn't make any difference (at least during my test cases), is there any reason why was it added? maybe I am missing some case and this is useful to avoid issues?

Controller/Adminhtml/Menu/Save.php
```diff
@@ -194,9 +194,9 @@ public function execute()
 
                     $nodeObject->setMenuId($menu->getMenuId());
                     $nodeObject->setTitle($node['title']);
-                    $nodeObject->setIsActive((int) ($node['is_active'] ?? 0));
-                    $nodeObject->setLevel($level);
-                    $nodeObject->setPosition($position);
+                    $nodeObject->setIsActive($node['is_active'] ?? '0');
```
jochhop, line 197: I can see that `is_active` methods are declared to receive `int` type as parameter and returns `int` as well, not sure why was this changed to be `string`? (applies to level and position)

Model/Menu.php
akai-z: I added this in order to avoid running unnecessary `delete` SQL query and the `for` loop `insert` queries (which can be refactored into `insertMultiple`) when there are no store changes.  
I also added it to avoid invalidating page cache [here](https://github.com/SnowdogApps/magento2-menu/blob/837ce8d2ed355bc8e19f0b61e595742b10b503db/Model/Menu.php#L79) unnecessarily, when there are no stores data changes.

The only downside to it is that if there are actually store changes, then it will run an additional `select` SQL query beside the delete and insert queries.  
However, I think it's worth it, as we should not delete and reinsert unchanged stores records,  
whenever we are just editing menu nodes for example (Which is probably the most thing to do when editing a menu).

Model/Menu/NodeRepository.php
akai-z: I added this because when Magento [loads a collection](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Data/Collection/AbstractDb.php#L552-L566) and [adds items data to it](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Data/Collection/AbstractDb.php#L588),  
the items that get added to collection are [initially empty](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Data/Collection/AbstractDb.php#L584).

When [items `addData`](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/DataObject.php#L44-L58) happens, and it [sets data to empty items objects](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Model/AbstractModel.php#L359),  
[Magento will treat the objects as objects that have data changes](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Model/AbstractModel.php#L356-L358), because `array_key_exists($key, $this->_data)`  
will be `false` as `$this->_data` is empty and getting filled with data.

Thus, when a [collection list](https://github.com/SnowdogApps/magento2-menu/blob/837ce8d2ed355bc8e19f0b61e595742b10b503db/Controller/Adminhtml/Menu/Save.php#L122) items are used to [save data](https://github.com/SnowdogApps/magento2-menu/blob/837ce8d2ed355bc8e19f0b61e595742b10b503db/Controller/Adminhtml/Menu/Save.php#L201), Magento will treat them as [modified](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Model/ResourceModel/Db/AbstractDb.php#L405-L410),  
and will [update](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Model/ResourceModel/Db/AbstractDb.php#L418-L420) them even without data changes.

That also means that the page cache will get unnecessarily invalidated.

I could move `$objectModel->setHasDataChanges(false);` outside the repository `getList` method and add it [here](https://github.com/SnowdogApps/magento2-menu/blob/837ce8d2ed355bc8e19f0b61e595742b10b503db/Controller/Adminhtml/Menu/Save.php#L126-L128) instead.

Model/MenuRepository.php
akai-z: https://github.com/SnowdogApps/magento2-menu/pull/128#discussion_r487841317

Right now, only the [nodes repository collection items has the save issue](https://github.com/SnowdogApps/magento2-menu/blob/837ce8d2ed355bc8e19f0b61e595742b10b503db/Controller/Adminhtml/Menu/Save.php#L201), so I could remove this line from menu repository.

Controller/Adminhtml/Menu/Save.php
akai-z: I made this change because when Magento loads menu and node objects, the `integer` data inside them get loaded as `string` (even if they were initially stored as integers).  
And when Magento tries to set `integer` data inside those object attributes, it will label the object as an [object with data changes](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Model/AbstractModel.php#L356-L358).  
That's because `$this->_data[$key] !== $value` will be `true`, due to data type differences, even if the values are the same.

Thus, when Magento tries to save the object, it will treat it as [modified](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Model/ResourceModel/Db/AbstractDb.php#L405-L410), and will it [update](https://github.com/magento/magento2/blob/2.4.0/lib/internal/Magento/Framework/Model/ResourceModel/Db/AbstractDb.php#L418-L420) it even without actual data changes.
That also means that the page cache will get unnecessarily invalidated.

I think this is currently the simplest workaround available to the issue, since that the attributes are not type hinted.

Model/Menu/NodeRepository.php
jochhop: I have tested it and indeed, if I remove the line, cache gets invalidated even if no changes are added to the menu. Thanks for the detailed explanation, setting this comment as resolved.

Model/MenuRepository.php
jochhop: Indeed only node repository has the issue, therefore, please remove this line, thanks in advance

Model/Menu.php
jochhop: Ok, it sounds logic and I think we can keep it, thanks

Controller/Adminhtml/Menu/Save.php
jochhop: You're right, I have tested a few cases and indeed this workaround resolves the issue

Model/Menu/NodeRepository.php
```diff
@@ -44,7 +51,9 @@ public function __construct(
     public function save(NodeInterface $object)
     {
         try {
+            $isPageCacheClearable = $object->hasDataChanges();
```
jochhop, line 54: What do you think if we implement cache invalidation through plugins? this way we could decouple the code so external extensions can get rid of this feature disabling them [plugins]

I can see Magento core implements it this way, eg: https://github.com/magento/magento2/blob/2.4.0/app/code/Magento/Catalog/Plugin/Model/ResourceModel/Attribute/Save.php#L44

I think this applies to all the methods which invalidates cache (`save` and `delete` in menus and nodes repositories)

akai-z: The plugins are now implemented.

Model/MenuRepository.php
akai-z: I removed this one.

Model/Menu.php
```diff
@@ -44,6 +76,7 @@ public function saveStores(array $stores)
             $connection->insert($table, ['menu_id' => $this->getId(), 'store_id' => $store]);
         }
         $connection->commit();
+        $this->cache->invalidatePageCache();
```
jochhop, line 79: I think we can also refactor this line so it's invalidated through a plugin?

akai-z: The plugin for this one is now implemented.

jochhop (approved):

jochhop (dismissed):

KamilBalwierz (approved):

jochhop (dismissed):

jochhop (approved):

PR #129, opened 2020-09-11
[69084] Add option to enable/disable menu

TM18:


Ui/Component/Listing/Column/MenuList/PageActions.php
```diff
@@ -10,17 +10,23 @@ public function prepareDataSource(array $dataSource)
         if (isset($dataSource["data"]["items"])) {
             foreach ($dataSource["data"]["items"] as & $item) {
                 $name = $this->getData("name");
-                $id = "X";
-                if (isset($item["menu_id"])) {
-                    $id = $item["menu_id"];
-                }
+                $id = $item["menu_id"] ?? "X";
```
jochhop, line 13: I also like using the `??` but I have seen that in main composer.json, minimum requirement for the menu module is: `"magento/framework": "100.1.*|101.0.*|102.0.*|103.0.*"`, which is Magento 2.1.x, still compatible with php 5.6, https://github.com/magento/magento2/blob/2.1/composer.json#L11

So I think we should add a constraint in snowdog menu's composer.json to restrict PHP version to be equal or greater than 7.x

jochhop (dismissed):

Controller/Adminhtml/AbstractMenu.php
```diff
@@ -0,0 +1,36 @@
...
+     * Authorization level of a basic admin session
+     *
+     * @see _isAllowed()
+     */
+    const ADMIN_RESOURCE = 'Snowdog_Menu::menus';
+
+    /**
+     * Menu ID key
+     */
+    const ID = 'menu_id';
```
mcjwsk, line 22: Is it really necessary?

Controller/Adminhtml/Menu/Edit.php
```diff
@@ -1,58 +1,100 @@
...
-            return $result;
-        } catch (NoSuchEntityException $e) {
-            $result = $this->resultRedirectFactory->create();
-            $result->setPath('*/*/index');
-            return $result;
+        $menuId = (int) $this->getRequest()->getParam(self::ID);
+
+        if ($menuId) {
+            try {
+                $model = $this->menuRepository->getById($menuId);
```
mcjwsk, line 75: Maybe we should move this part to `\Snowdog\Menu\Controller\Adminhtml\AbstractMenu` as `initMenu` method? I see it is used in other places too.

Controller/Adminhtml/Menu/Save.php
```diff
@@ -249,16 +260,20 @@ private function validateProductNode(array $node)
...
     {
-        $menuId = $this->getRequest()->getParam('id');
+        $menuId = (int) $this->getRequest()->getParam(self::ID);
 
         if ($menuId) {
-            return $this->menuRepository->getById($menuId);
+            try {
+                return $this->menuRepository->getById($menuId);
+            } catch (NoSuchEntityException $exception) {
+                return $this->menuFactory->create();
```
mcjwsk, line 275: are you sure?

Model/MenuRepository.php
```diff
@@ -149,4 +150,16 @@ public function get($identifier, $storeId)
         $collection->addFilter('store_id', $storeId);
         return $collection->getFirstItem();
     }
+
+    public function setIsActiveByIds($ids, $isActive)
+    {
+        /** @var Collection $collection */
+        $collection = $this->collectionFactory->create();
+        $connection = $collection->getConnection();
+        $connection->update(
```
mcjwsk, line 159: this should be a part of menu resource model

Ui/Component/Listing/Column/Stores.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Snowdog\Menu\Ui\Component\Listing\Column;
+
+use Magento\Store\Ui\Component\Listing\Column\Store\Options as StoreOptions;
+
+class Stores extends StoreOptions
```
mcjwsk, line 9: are you sure there is no method for that in magento core? if I understand it correctly you want to add `All Store Views` option, right?

Ui/DataProvider/Menu/Form/MenuDataProvider.php
```diff
@@ -0,0 +1,49 @@
...
+            $primaryFieldName,
+            $requestFieldName,
+            $meta,
+            $data
+        );
+    }
+
+    public function getData(): array
+    {
+        if (!empty($this->loadedData)) {
```
mcjwsk, line 36: do not use `empty()` to check field is initialized, use `$this->loadedData !== null` instead (remove default property value of course at first)

Ui/DataProvider/Menu/Form/MenuDataProvider.php
```diff
@@ -0,0 +1,49 @@
...
+            return $this->loadedData;
+        }
+
+        $items = $this->collection->getItems();
+        /** @var Menu $menu */
+        foreach ($items as $menu) {
+            $menu->addData(['stores' => $menu->getStores()]);
```
mcjwsk, line 43: maybe we should add `addStoresData()` method in collection class?

view/adminhtml/ui_component/snowmenu_menu_list.xml
```diff
@@ -36,6 +36,29 @@
...
     </dataSource>
+    <listingToolbar name="listing_top">
+        <paging name="listing_paging"/>
+        <massaction name="listing_massaction">
+            <action name="enable">
+                <argument name="data" xsi:type="array">
+                    <item name="config" xsi:type="array">
+                        <item name="type" xsi:type="string">enable</item>
+                        <item name="label" translate="true" xsi:type="string">Enable</item>
+                        <item name="url" xsi:type="url" path="snowmenu/menu/massEnable/"/>
```
mcjwsk, line 47: I am not sure it makes a difference but you have extra slash at the end of path.

Model/ResourceModel/Menu/Collection.php
```diff
@@ -5,6 +5,8 @@
 
 class Collection extends AbstractCollection
 {
+    protected $_idFieldName = 'menu_id';
```
mcjwsk, line 8: Why is it necessary?

Controller/Adminhtml/Menu/Edit.php
```diff
@@ -1,58 +1,100 @@
...
+        $this->coreRegistry->register(self::REGISTRY_CODE, $model);
+
+        /** @var Page $resultPage */
+        $resultPage = $this->resultPageFactory->create();
+        $this->initPage($resultPage)->addBreadcrumb(
+            $menuId ? __('Edit Menu %1', $model->getTitle()) : __('New Menu'),
+            $menuId ? __('Edit Menu %1', $model->getTitle()) : __('New Menu')
+        );
+        $resultPage->getConfig()->getTitle()->prepend(__('Menus'));
+        $resultPage->getConfig()->getTitle()->prepend($model->getId() ? $model->getTitle() : __('New Menu'));
```
mcjwsk, line 96: extract `$resultPage->getConfig()->getTitle()` to some variable because you use it twice

Controller/Adminhtml/Menu/Edit.php
```diff
@@ -1,58 +1,100 @@
...
+            $context
+        );
     }
 
     /**
-     * Dispatch request
-     *
-     * @return \Magento\Framework\Controller\ResultInterface|ResponseInterface
-     * @throws \Magento\Framework\Exception\NotFoundException
+     * @SuppressWarnings(PHPMD.NPathComplexity)
```
mcjwsk, line 67: I think you should refactor the method to avoid this statement

Controller/Adminhtml/AbstractMenu.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Snowdog\Menu\Controller\Adminhtml;
+
+use Magento\Backend\App\Action;
+use Magento\Backend\Model\View\Result\Page;
+
+abstract class AbstractMenu extends Action
```
mcjwsk, line 10: What do you think of `MenuAction` instead of `AbstractMenu` ?

mcjwsk (changes requested):
🦆

Ui/Component/Listing/Column/Stores.php
TM18: For the previous view I don't think there was such method. Now that this field is changed to `All Store Views` style, there is a core Magento method so I removed this file and use the core one.

Ui/DataProvider/Menu/Form/MenuDataProvider.php
TM18: Changed

Ui/DataProvider/Menu/Form/MenuDataProvider.php
TM18: Good idea, added this method to the collection class.

view/adminhtml/ui_component/snowmenu_menu_list.xml
TM18: It does not but I removed it anyway so it's consistent with other urls.

Model/ResourceModel/Menu/Collection.php
TM18: It was before when I was testing other solution, now I guess it's not necessary anymore so I deleted it.

Model/MenuRepository.php
TM18: As discussed, we'll keep it like this as collection is still part of a resource model and here changes are made to a collection of menus.

Controller/Adminhtml/Menu/Edit.php
kiber: Done

Controller/Adminhtml/AbstractMenu.php
kiber: Done

Controller/Adminhtml/Menu/Save.php
kiber: Why not? If there is no menu id, it creates a new menu model for saving data.

Controller/Adminhtml/AbstractMenu.php
kiber: Yes, it is necessary. We will use it in child classes where we set `id` instead of `menu_id`.

Controller/Adminhtml/Menu/Edit.php
kiber: How did you suppose to move it to initMenu method?

Controller/Adminhtml/Menu/Edit.php
kiber: I have tried, but there is no setTitle() method to set it after declaring the getTitle() method as a variable. That is why I left it as it.

Controller/Adminhtml/Menu/Edit.php
kiber: Removed it. Seems we don't need it.

Controller/Adminhtml/Menu/Edit.php
jochhop: I think Darek meant to add it to an abstract class, so it can be called from there in children classes, something like this example within Magento 2 core: 

Method definition: https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Customer/Controller/Adminhtml/Index.php#L259
Method call: https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Customer/Controller/Adminhtml/Index/Cart.php#L157

Controller/Adminhtml/Menu/Edit.php
jochhop: You're right Samir, I have checked `getTitle()` method and any time is called, it builds the page through layouts preparation, therefore, if assigned to a variable, it'll not prepend the title

Controller/Adminhtml/Menu/Status.php
```diff
@@ -0,0 +1,70 @@
...
+
+        return $redirect;
+    }
+
+    /**
+     * Returns menu model based on the Request (requested with `id` or fresh instance)
+     *
+     * @return Menu
+     */
+    private function getCurrentMenu(): Menu
```
jochhop, line 60: I see this method repeated at least in another class: https://github.com/SnowdogApps/magento2-menu/pull/129/files#diff-3d71cb48bdc8ebcb5edfa93998475c39edccc1d9fe4958d1aa8cdd3939e0283eL256

Maybe it's worth to move it to some abstract class and call it there instead

TM18: I moved it to abstract class `MenuAction`.

Controller/Adminhtml/Menu/Edit.php
TM18: I used `getCurrentMenu()` method here as init method would do basically the same.

Block/Adminhtml/Button/AbstractButton.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
```
jochhop, line 2: We can add already `declare(strict_types=1);` in classes, we started using it already, eg: https://github.com/SnowdogApps/magento2-menu/blob/develop/Block/NodeType/Product.php (we'll need to drop php 5.x compatibility en composer.json soon)

Block/Adminhtml/Button/AbstractButton.php
```diff
@@ -0,0 +1,26 @@
...
+{
+    /** @var Context */
+    protected $context;
+
+    public function __construct(Context $context)
+    {
+        $this->context = $context;
+    }
+
+    protected function getUrl($route, $params = []): string
```
jochhop, line 17: i guess we can start using strong types (so we could start adding it in all new classes [params and method return types])?

Controller/Adminhtml/Menu/Edit.php
```diff
@@ -1,58 +1,80 @@
...
 
-    protected function _isAllowed()
-    {
-        return $this->_authorization->isAllowed('Snowdog_Menu::menus');
+        $this->coreRegistry->register(self::REGISTRY_CODE, $menu);
+
+        /** @var Page $resultPage */
+        $resultPage = $this->resultPageFactory->create();
+        $this->initPage($resultPage)->addBreadcrumb(
+            $menuId ? __('Edit Menu %1', $menu->getTitle()) : __('New Menu'),
```
jochhop, line 72: This is repeated, maybe we can create a variable and store it and use it in both placeS?

Service/MenuHydrator.php
```diff
@@ -17,9 +17,11 @@ class MenuHydrator
      */
     public function mapRequest(MenuInterface $menu, RequestInterface $request): MenuInterface
     {
-        $menu->setTitle($request->getParam('title'));
-        $menu->setIdentifier($request->getParam('identifier'));
-        $menu->setCssClass($request->getParam('css_class'));
+        $params = $request->getParams();
+        $menu->setTitle($params['title']);
```
jochhop, line 21: I think here and below lines we could use references from `MenuInterface` constants like: `MenuInterface::TITLE`

Ui/Component/Listing/Column/MenuList/PageActions.php
```diff
@@ -112,4 +115,24 @@ private function getDeleteButton(int $menuId): array
             '__disableTmpl' => true,
         ];
     }
+
+    /**
+     * @param int $menuId
+     * @param bool $isActive
+     * @return array
+     */
```
jochhop, line 123: I think these docs is a duplicated info and we could remove it (params and return type are already defined within method declaration)

Controller/Adminhtml/Menu/Edit.php
```diff
@@ -1,58 +1,80 @@
...
-        parent::__construct($context);
-        $this->menuRepository = $menuRepository;
-        $this->registry = $registry;
+    /**
+     * @param Context $context
+     * @param Registry $coreRegistry
+     * @param PageFactory $resultPageFactory
+     * @param MenuRepositoryInterface $menuRepository
+     * @param MenuFactory $menuFactory
+     */
```
jochhop, line 38: I think we can remove `__construct` docs as all params types are already defined within the method declaration

Block/Adminhtml/Button/AbstractButton.php
TM18: Ok, added `declare(strict_types=1);`

Block/Adminhtml/Button/AbstractButton.php
TM18: Sure, added strong types to params.

Controller/Adminhtml/Menu/Edit.php
TM18: Extracted to a variable.

Service/MenuHydrator.php
TM18: Good idea, replaced with `MenuInterface` references.

Ui/Component/Listing/Column/MenuList/PageActions.php
TM18: Right, removed them.

Controller/Adminhtml/Menu/Edit.php
TM18: Done.

JKrupinski (approved):

emastyle (approved):

akai-z (changes requested):
TODO: Fix [develop branch merge](https://github.com/SnowdogApps/magento2-menu/pull/129/commits/c612c2d1e76dded50c394a2cf560746858d20586) frontend conflicts in the following files:
- [`README.md`](https://github.com/SnowdogApps/magento2-menu/pull/129/commits/c612c2d1e76dded50c394a2cf560746858d20586#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R174-R178) (The conflict incoming change from `develop` branch is the correct one, but there is an additional comma after the word `root` in the current feature branch, and I wanted to check whether we still need it.)
- [`view/adminhtml/web/vue/app.vue`](https://github.com/SnowdogApps/magento2-menu/pull/129/commits/c612c2d1e76dded50c394a2cf560746858d20586#diff-8dbe788dbbce257cc6b22b041d535da6c8766c2e4d82117cb5e5cee25e71eb2aR119-R130)
- [`view/adminhtml/web/vue/field-type/checkbox.vue`](https://github.com/SnowdogApps/magento2-menu/pull/129/commits/c612c2d1e76dded50c394a2cf560746858d20586#diff-935df278827c55ea4671982ce7d62a9b4e14e846498c3a02eb0006ad5b2f9bb1R17-R26)

akai-z (approved):

emastyle (approved):

Issue #95, opened 2019-10-02
Using a product node displays original image


tomkingfisher:
When using a product node it displays the full original image rather than a resized and optimised version.

Because the menu is on ever page the images are being loaded on every page and because they are the original images it is slowing the site down.

Is there a way to be able to choose an image role or specify a role so that images are optimised and resized?

kiber:
@tomkingfisher changes are merged with the **develop** branch and it will be in the next release.
You can use
`$resizedProductImage = $block->getResizedProductImage($nodeId, '300', '300');`
in template file to get resized product image

andrzejborkowski:
@tomkingfisher the issue has been resolved in the latest [2.13.0 release](https://github.com/SnowdogApps/magento2-menu/releases/tag/2.13.0)

PR #137, opened 2020-10-01
DEV-70199 added image helper to resize product image

kiber:
https://github.com/SnowdogApps/magento2-menu/issues/95


Block/NodeType/Product.php
```diff
@@ -105,13 +116,13 @@ public function fetchData(array $nodes)
     {
         $storeId = $this->_storeManager->getStore()->getId();
 
-        list(
+        [
```
jochhop, line 119: This is a feature which can be used since PHP 7.1: https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.symmetric-array-destructuring

However the composer.json of the module points as minimum requirement Magento Framework version `100.1.*` (https://github.com/SnowdogApps/magento2-menu/blob/develop/composer.json#L21), which can be compatible with PHP 5.6 and PHP 7.0, so I think we should keep the `list` way to avoid issues with compatibilities.

Block/NodeType/Product.php
```diff
@@ -178,6 +189,27 @@ public function getProductImage($nodeId)
         return $this->getMediaUrl('catalog/product' . $image);
     }
 
+    /**
+     * @param $nodeId
```
jochhop, line 193: Shouldn't be this typed as `int`?

Helper/Image.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * First check this file on FS
+     *
+     * @param string $filename
+     * @return bool
+     */
+    private function fileExists(string $filename): bool
+    {
+        if ($this->mediaDirectory->isFile($filename)) {
```
jochhop, line 112: `$this->mediaDirectory->isFile($filename)` already returns a boolean so we can skip the condition and just return its value?

Helper/Image.php
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @param string $image
+     * @param string|null $width
+     * @param string|null $height
+     * @return string|null
+     * @throws NoSuchEntityException
+     */
+    public function resize(string $image, $width = null, $height = null): ?string
+    {
+        $path = self::DIRECTORY;
```
jochhop, line 70: Please extract `$path` creation into a separated private method

Helper/Image.php
```diff
@@ -0,0 +1,118 @@
...
+            $imageFactory->open($absoluteImagePath);
+            $imageFactory->constrainOnly(true);
+            $imageFactory->keepTransparency(true);
+            $imageFactory->keepFrame(true);
+            $imageFactory->keepAspectRatio(true);
+            $imageFactory->resize($width, $height);
+            $imageFactory->save($resizedImagePath);
+        }
+
+        $resizedImage = $this->storeManager->getStore()
```
jochhop, line 96: not sure what you think, but this format seems to me weird, concatenating a method call with strings seems confusing, but this is just a suggestion, to better readable code, I'd refactor to this:

```php
$resizedImage = $this->storeManager->getStore()
                ->getBaseUrl(UrlInterface::URL_TYPE_MEDIA);
$resizedImage .= $path . $image;
```

jochhop (changes requested):

jochhop (approved):

KamilBalwierz (approved):

Issue #75, opened 2019-01-29
Include subcategories


DanieliMi:
Hey,
is it possible to implement a functionality to import categories including subcategories as nodes? Maybe even with specifying until which level you want to include subcategories. It would be really handy if you could import them in case you have a whole bunch of subcategories. Inserting all these nodes separately is lots of effort.

jg-development:
+1

digiltd:
@jg-development this already works, add the "category child" node as the child of a regular category node (note: I don't think this is a requirement, but should help you get your head around it)

jg-development:
ahh thx .. will try

template-provider:
for me it works

andy1786:
for me works, only the ordering with drag and drop is not working

andrzejborkowski:
Import categories and subcategories as menu nodes was released recently in [2.16.0](https://github.com/SnowdogApps/magento2-menu/releases/tag/2.16.0)

PR #138, opened 2020-10-01
[70196] added import nodes from categories tree

kiber:
https://github.com/SnowdogApps/magento2-menu/issues/75


view/adminhtml/web/js/mixins/modal-mixin.js
```diff
@@ -0,0 +1,64 @@
...
+                  dataType: 'json',
+                  success: function (data) {
+                    vueApp.list = data.list
+
+                    alertModal.closeModal(true)
+                    importModal.closeModal()
+                  },
+                  error: function (result) {
+                    alertModal.closeModal(true)
+                  }
```
anqaka, line 55: shouldn't we display error message to notify customer that error occured?

JKrupinski: Yes, I've added alert with error message, thanks

anqaka: 👍

anqaka (approved):

PR #140, opened 2020-10-08
[70218] Add menu node image field

### LINKED ISSUE #114: Image upload option for node
It would be nice to have the option to upload an image to a node. A use case for this would be for a mega menu which contains promotional banners.

This can already be achieved using Cms Block node type, but it can get quite messy to manage, for example if you have a menu with 10 top-level categories, and you want two or three promotional banners within each one. It would be far nicer from a user/content-management perspective if the image could be uploaded directly to the node to avoid having to create separate static blocks.

akai-z:
Resolves #114


akai-z (commented):
TODO: Fix [develop branch merge](https://github.com/SnowdogApps/magento2-menu/pull/140/commits/90a4870bdb588c6b6f5bc7e9be8b89f9bb338101) conflicts in the following files:
[`view/adminhtml/templates/menu/nodes.phtml`](https://github.com/SnowdogApps/magento2-menu/pull/140/commits/90a4870bdb588c6b6f5bc7e9be8b89f9bb338101#diff-9f49d9df8801512b198d845841fc4add2a48a2b43c4c5c3858eb67847466b900)
[`view/adminhtml/web/vue/menu-type.vue`](https://github.com/SnowdogApps/magento2-menu/pull/140/commits/90a4870bdb588c6b6f5bc7e9be8b89f9bb338101#diff-cbb716bce7b489abd95939ddc4983de82ee912bec7f876b2d3e55099a22ec5db)

emastyle (commented):
The backend part looks good to me. Vue part still needs to CR by frontend dev and if ok all can be approved.

Block/Menu/Edit/Tab/Nodes.php
```diff
@@ -118,6 +125,30 @@ public function isHidden()
...
+        return $this->getUrl('snowmenu/node/uploadimage');
+    }
+
+    /**
+     * @return string
+     */
+    public function getImageDeleteUrl()
+    {
+        return $this->getUrl('snowmenu/node/deleteimage');
+    }
```
emastyle, line 142: What do you think if we set the URLs in constants? I.e.:

`const IMAGE_UPLOAD_URL = 'snowmenu/node/uploadimage';`

then in the method:

`public function getImageUploadUrl()
   {
        return $this->getUrl(self::IMAGE_UPLOAD_URL);
   }`

emastyle (commented):
The backend code is ok for me, please just check the comment I left for Block/Menu/Edit/Tab/Nodes.php.

Block/Menu/Edit/Tab/Nodes.php
akai-z: Yeah, I updated it.

jochhop (approved):

PR #144, opened 2020-10-26
[70197] Menu YAML import/export feature

### LINKED ISSUE #82: Import/export nodes or putting them under versionning
Is there a way to import/export menus and nodes? Wondering what kind of deployment strategies there is when using this module. Can we load nodes from a file, put them under git versioning maybe?

akai-z:
Resolves #82

### DIFF
--- a/Controller/Adminhtml/Menu/DownloadImportSample.php
+++ b/Controller/Adminhtml/Menu/DownloadImportSample.php
@@ -66,6 +66,6 @@ class DownloadImportSample extends Action implements HttpGetActionInterface
 
         $resultRedirect = $this->resultRedirectFactory->create();
 
-        return $resultRedirect->setPath('*/*');
+        return $resultRedirect->setPath('*/*/import');
     }
 }
--- a/Controller/Adminhtml/Menu/DownloadImportSample.php
+++ b/Controller/Adminhtml/Menu/DownloadImportSample.php
@@ -6,6 +6,7 @@ namespace Snowdog\Menu\Controller\Adminhtml\Menu;
 
 use Exception;
 use Magento\Backend\App\Action;
+use Magento\Backend\App\Action\Context;
 use Magento\Framework\App\Action\HttpGetActionInterface;
 use Psr\Log\LoggerInterface;
 use Snowdog\Menu\Model\ImportExport\File\Download as FileDownload;
@@ -34,7 +35,7 @@ class DownloadImportSample extends Action implements HttpGetActionInterface
     private $sampleData;
 
     public function __construct(
-        Action\Context $context,
+        Context $context,
         LoggerInterface $logger,
         FileDownload $fileDownload,
         SampleData $sampleData
--- a/Controller/Adminhtml/Menu/DownloadImportSample.php
+++ b/Controller/Adminhtml/Menu/DownloadImportSample.php
@@ -55,7 +55,7 @@ class DownloadImportSample extends Action implements HttpGetActionInterface
         try {
             return $this->fileDownload->generateDownloadFile(
                 SampleData::DOWNLOAD_FILE_ID,
-                $this->sampleData->getSampleData()
+                $this->sampleData->get()
             );
         } catch (Exception $exception) {
             $this->logger->critical($exception);
--- a/Controller/Adminhtml/Menu/DownloadImportSample.php
+++ b/Controller/Adminhtml/Menu/DownloadImportSample.php
@@ -8,7 +8,7 @@ use Exception;
 use Magento\Backend\App\Action;
 use Magento\Framework\App\Action\HttpGetActionInterface;
 use Psr\Log\LoggerInterface;
-use Snowdog\Menu\Model\ImportExport\FileDownload;
+use Snowdog\Menu\Model\ImportExport\File\Download as FileDownload;
 use Snowdog\Menu\Model\ImportExport\Import\SampleData;
 
 class DownloadImportSample extends Action implements HttpGetActionInterface
--- a/Controller/Adminhtml/Menu/DownloadImportSample.php
+++ b/Controller/Adminhtml/Menu/DownloadImportSample.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Snowdog\Menu\Controller\Adminhtml\Menu;
 
+use Exception;
 use Magento\Backend\App\Action;
 use Magento\Framework\App\Action\HttpGetActionInterface;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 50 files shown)


Model/Menu/ExportProcessor.php
```diff
@@ -0,0 +1,169 @@
...
+        $searchCriteria = $this->searchCriteriaBuilder
+            ->addFilter(NodeInterface::MENU_ID, $menuId)
+            ->create();
+
+        $nodes = $this->nodeRepository->getList($searchCriteria)->getItems();
+        $nodesData = [];
+
+        foreach ($nodes as $key => $node) {
+            $nodesData[$key] = $node->getData();
+            unset($nodesData[$key][NodeInterface::MENU_ID]);
```
jochhop, line 137: What do you think if we avoid to export node ID as well? I see later, when importing, it's removed, but I think we don't need it in export file either: https://github.com/SnowdogApps/magento2-menu/pull/144/files#diff-64fbf188d4c313434d60faecd6839110c25242e57597f3ca1c7bf073434a9340R125

akai-z: Node ID alongside Parent ID are important in order identify the parent-child relations between nodes.  
(i.e. If a node parent ID matches another node ID, then it's a child of that node.)

I'll apply some changes to the code to make importing node ID and parent ID optional,  
for nodes that don't have children and nodes that don't have parents.

Model/Menu/ImportProcessor.php
```diff
@@ -0,0 +1,94 @@
...
+        $data = $this->uploadFileAndGetData();
+        $menu = $this->createMenu($data);
+
+        $this->nodeImportProcessor->createNodes($data[ExportProcessor::NODES_CSV_FIELD], $menu->getId());
+
+        return $menu->getIdentifier();
+    }
+
+    /**
+     * @return \Snowdog\Menu\Api\Data\MenuInterface
```
jochhop, line 52: Could you please import this Interface reference so it matches the rest of the code style?

akai-z: We decided not change this one as `MenuInterface` is not used in the class.

jochhop (dismissed):

Controller/Adminhtml/Menu/DownloadImportSample.php
```diff
@@ -0,0 +1,69 @@
...
+        $this->logger = $logger;
+        $this->importSampleFile = $importSampleFile;
+
+        parent::__construct($context);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute()
```
KamilBalwierz, line 51: I do not see that we provide any sample file (nor create one). Also it looks like it should be in {magentoRoot}/var and as far as I know we do not have to move file there via composer.json instructions

Controller/Adminhtml/Menu/Export.php
```diff
@@ -0,0 +1,71 @@
...
+        $this->logger = $logger;
+        $this->exportProcessor = $exportProcessor;
+
+        parent::__construct($context);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute()
```
KamilBalwierz, line 51: Do we really want to store all exports as downloadable files?
This causes:
* mess in `var`
* potential problems with writing files and permissions
* pseudorandom suffix in file name may be not sufficient for providing uniqueness

Model/Menu/ExportProcessor.php
KamilBalwierz: OR replace Node IDs with some UUID that are vaild only during export/import to avoid using those in database

OR use some better data format - originally when thinking about this functionality I was envisiong using YAML or JSON so structure can be implied by data format (in theory YAML would be more user friendly allowing manual edditing in almost-human-readable format)

Controller/Adminhtml/Menu/DownloadImportSample.php
KamilBalwierz: ok, found code that generates sample file ...

Model/Menu/ImportProcessor/Menu.php
```diff
@@ -0,0 +1,114 @@
...
+    private function getNewMenuIdentifier($identifier)
+    {
+        $menus = $this->getMenuListByIdentifier($identifier);
+        $identifiers = [];
+
+        foreach ($menus as $menu) {
+            $identifiers[$menu->getIdentifier()] = $menu->getId();
+        }
+
+        while (isset($identifiers[$identifier])) {
```
KamilBalwierz, line 95: so this for mutiple import of the same stuff will create:
mymenu
mymenu-1
mymenu-1-1
mymenu-1-1-1
mymenu-1-1-1-1

maybe we should think about making it
mymenu
mymenu-1
mymenu-2
and so on ? 😄

akai-z: This is actually a similar behavior in Magento 2 products SKUs, if you try to keep duplicating already duplicated products.

I applied some changes to improve it by incrementing the last part of a menu identifier.  
I made the identifier provided in CSV import file to always be the base for the new identifier, even if it already contains hyphens with numeric values.

If we try to dissect the import identifiers that already have hyphens further, which would be a bit more complicated,  
then we have to make sure that all identifiers with hyphens are fully/properly loaded from DB to avoid creating duplicate identifiers and avoid any unnecessary additional DB checks when generating a new identifier.

Note:  
In Magento 2 CMS module, pages and blocks duplicates' identifiers get generated using [`uniqid()`](https://github.com/magento/magento2/blob/2.4.1/app/code/Magento/Cms/Controller/Adminhtml/Block/Save.php#L129).  
But the requirements for this menu import feature was to use Magento 2 products duplication approach,  
and append a numeric value at the end of the identifier.

Controller/Adminhtml/Menu/Export.php
akai-z: Export files do not get permanently stored, they get temporarily stored to obtain the download file content.  
Before the download file response is returned, the file will get deleted. ([Check HTTP File Response `rm` flag](https://github.com/SnowdogApps/magento2-menu/blob/cc67abd08368843130b186fe40fc67548a70fcdd/Model/Menu/ExportProcessor.php#L100))  
That's how Magento HTTP file download response class [works](https://github.com/magento/magento2/blob/2.4.1/lib/internal/Magento/Framework/App/Response/Http/FileFactory.php#L107-L109).

I opted to use the ready [Magento file download](https://github.com/magento/magento2/blob/2.4.1/lib/internal/Magento/Framework/App/Response/Http/FileFactory.php) solution rather than writing the file download logic in the module,  
to keep the code as simple as possible, but with the caveat that export files will get temporarily stored before being downloaded.

The same goes for the menu import processor, which uses Magento ImportExport module [CSV file upload and data processing](https://github.com/magento/magento2/blob/2.4.1/app/code/Magento/ImportExport/Model/Import.php#L612-L630) feature.  
The temporary import file [gets deleted](https://github.com/SnowdogApps/magento2-menu/blob/cc67abd08368843130b186fe40fc67548a70fcdd/Model/Menu/ImportProcessor.php#L88) after obtaining the processed upload file data.

jochhop (dismissed):

emastyle (approved):
Code looks good to me, also testing in local went fine without problems.

jochhop (approved):

Issue #151, opened 2020-12-24
Enabled Checkbox warning regarding Invalid prop  type


vmalyk:
Pre-condition:
Snowdog module version: 2.13.0

On check/uncheck checkbox We got warning regarding prop type for checkbox:
<img width="952" alt="Снимок экрана 2020-12-24 в 13 36 40" src="https://user-images.githubusercontent.com/5318027/103083305-1eb9f080-45ed-11eb-85c8-a80f85c6f55e.png">

Issue related with checkboxValue watcher and check returns Boolean instead declared Sting or Number.

PR #153, opened 2020-12-24
Fix warning about type

vmalyk:
Fix for #151.


Igloczek (approved):

Igloczek:
Thanks for spotting it out and fixing 🙇

Igloczek:
BTW. You don't need to open the issue and PR separately, the same description as in the issue, but in PR is enough.

PR #157, opened 2021-01-24
[74936] Fix custom templates for node types wrapper and custom_url

thampe:
Hi,

first of all thanks for providing this extension!

This PR provides a fix for the custom template features on the node types `wrapper` and `custom_url`.

### DIFF
--- a/Block/NodeType/CustomUrl.php
+++ b/Block/NodeType/CustomUrl.php
@@ -58,9 +58,7 @@ class CustomUrl extends AbstractNode
      */
     public function getJsonConfig()
     {
-        $data =  $this->_customUrlModel->fetchConfigData();
-
-        return $data;
+        return $this->_customUrlModel->fetchConfigData();
     }
 
     /**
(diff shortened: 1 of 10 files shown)


andrzejborkowski:
Thank you for reporting and fixing the issue, we'll review your PR soon

akai-z:
I checked the fix and it worked properly.

But I found that `snowMenuSimpleField` data was not used anywhere in Vue.js files, thus it did not seem necessary.  
We need a frontend developer to verify that before removing it.

(I applied some minor code changes.)

akai-z (commented):
I checked the fix and it worked properly. (I applied some minor code changes.)

But I found that `snowMenuSimpleField` data was not used anywhere in Vue.js files, thus it did not seem necessary.  
We need a frontend developer to verify that before removing it.

akai-z (changes requested):
I found that `snowMenuSimpleField` data (which existed before the fix in this pull request) was not used anywhere in Vue.js files,  
thus it did not seem necessary.

We need a frontend developer to verify that, before removing it.

akai-z (dismissed):

akai-z (approved):

PR #172, opened 2021-04-15
Commit two ugly files to check CS from fork

emastyle:


emastyle:
We are closing this as it seems related to action/checkout@v2 (we are using in our workflow) issue that should be fixed in the next release.

Issue #161, opened 2021-03-01
[feature request] Any plans for GraphQL endpoints?


tiberiur:
Hello,

Snowdog_Menu is my go-to menu extension and I use it in almost every project.
Quick question: any plans for GraphQL endpoints?

Thanks for the great work so far!

andrzejborkowski:
hi @tiberiur! Thanks for your feedback, glad to hear you like our module! As for your request, we didn't work on this direction yet, but I created internal ticket DEV-76409 and let you know after we did some research

akai-z:
@tiberiur The GraphQL feature is done.  
(It's currently merged into `develop` branch, and will later be published in a new release.)  
If you have any questions about it, please let us know.

PR #175, opened 2021-04-20
Fix for #161

frezelth:
Issue with node type fetched instead of content in the graphQl nodes endpoint


Igloczek:
Thanks for spotting out and fixing this problem Thomas, fix will be released with next version 🙇

PR #181, opened 2021-05-17
Added Magento Yes/No toggle to checkbox component

jbclaudio:
I ran into the problem that the checkbox did not show whether it was enabled or disabled. (There was no cross in the checkbox).

Also, no use was made of the standard Yes / No checkbox of Magento.

I have changed the checkbox to the Yes / No checkbox and this works as a charm
image


JKrupinski (approved):

JKrupinski:
Thanks! Looks good to me

PR #188, opened 2021-06-11
[69080] Add creation time, update time and store view columns to menu admin grid and fix the mass delete action

akai-z:


akai-z:
The XML format that I used for the menu admin grid new columns in UI component `snowmenu_menu_list.xml` file is based on Magento 2.1.  
The reason I did that was to keep it consistent with the format used in the rest of file, which is based on Magento 2.1.

Here is an example of the XML format in Magento 2.1:  
https://github.com/magento/magento2/blob/2.1.0/app/code/Magento/Cms/view/adminhtml/ui_component/cms_block_listing.xml#L199-L208

And here is an example of the XML format in Magento 2.4:  
https://github.com/magento/magento2/blob/2.4.2/app/code/Magento/Cms/view/adminhtml/ui_component/cms_block_listing.xml#L174-L180

jochhop (approved):

KamilBalwierz (approved):

PR #198, opened 2021-07-27
[80889] Menu duplicate feature fixes and improvements

akai-z:


jochhop (approved):

Service/Menu/Cloner.php
```diff
@@ -79,30 +91,40 @@ public function clone(MenuInterface $menu): MenuInterface
...
+                $nodeClone->setMenuId($menuCloneId);
 
-            if ($node->getImage()) {
-                $nodeClone->setImage($this->nodeImage->clone($node->getImage()));
-            }
+                if (isset($nodeIdMap[$node->getParentId()])) {
+                    $nodeClone->setParentId($nodeIdMap[$node->getParentId()]);
+                }
+
+                if ($node->getImage()) {
```
KamilBalwierz, line 115: suggestion for further development: based on node type call `cloneNode` in node type handler class so every node type can have specific cloning instructions (like here possibly handle duplication of uploaded image for specific node)

KamilBalwierz (approved):

Model/NodeType/AbstractNode.php
```diff
@@ -34,14 +36,21 @@ abstract class AbstractNode implements NodeTypeInterface
...
+     * @var NodeImage
+     */
+    private $nodeImage;
+
     /**
      * AbstractNode constructor.
      */
     public function __construct(
-        \Magento\Framework\Profiler $profiler
+        \Magento\Framework\Profiler $profiler,
```
jochhop, line 48: Maybe let's import this class to keep the same code style

jochhop (approved):

PR #210, opened 2021-09-10
 NodeType CategoryChild: improve handling when child category is not defined or is a category that no longer exists

Echron:
In the adminhtml it's possible to not define a child category OR the child category can be removed. This code will prevent the frontend from crashing. Next step would also show some warning in the backend when saving a menu with invalid or empty children.


jochhop:
Hi @Echron , thanks for your contribution, could you please define detailed steps of how to replicate the issue? we're trying different paths but we cannot reproduce it in current develop branch, thanks in advance

andrzejborkowski:
@Echron please provide the steps how to replicate the issue, otherwise, we would have to close your PR. Thank you!

andrzejborkowski:
Since we didn't get any feedback and we can't replicate the issue, we have to close your PR. Please feel free to open it again if you still have this issue. Thank you!

PR #227, opened 2021-12-17
Add category url_key to graphql node interface

### LINKED ISSUE #225: GraphQL - Missing URL information for the Categories nodes
Sup guys,

I'm using this module for a PWA Studio project so that I'm also using the GraphQL APIs.

I created the logic to fetch the data, but we don't have a field that returns the url_key of the category.

My suggestion is to add a new field called **url_key** inside the node interface related to categories and Childs

texboy:
Resolves #225


etc/schema.graphqls
```diff
@@ -50,6 +50,7 @@ interface SnowdogMenuNodeInterface @typeResolver(class: "Snowdog\\Menu\\Model\\G
     submenu_template: String @doc(description: "Node submenu template")
     creation_time: String! @doc(description: "Node creation time")
     update_time: String! @doc(description: "Node update time")
+    url_key: String @doc(description: "Returns the category url key when available") @resolver(class: "Snowdog\\Menu\\Model\\GraphQl\\Resolver\\Node\\Field\\UrlKey")
```
jochhop, line 53: Hi @texboy , thanks a lot for your PR, but after discussing it internally, we think this is not the best way to implement this feature, the reasons are:

 * url_key should be available also for products and CMS pages, not only for categories
 * the current way loads single categories, therefore if a category is repeated in multiple nodes, it'll be loaded multiple times, therefore there'll be a performance issue in this case

We suggest the following:

* Instead of using a custom url_key field, we could use native resolvers for category/products/CMS page if a node uses one of these entities (the native resolves will provide url key and more additional data if needed)

Are you able to refactor this? 

Thanks again!

jochhop (changes requested):

etc/schema.graphqls
LucasCalazans: hey @jochhop, I think it's a valid point.
But just a reminder, for the **SnowdogMenuCustomUrlNode** type, the URL is already coming on the **content** field

PR #235, opened 2022-02-18
#223 Allow to set submenu template via layout xml

kiber:
DEV-87135

### DIFF
--- a/Block/Menu.php
+++ b/Block/Menu.php
@@ -99,8 +99,13 @@ class Menu extends Template implements DataObject\IdentityInterface
         $this->templateResolver = $templateResolver;
         $this->imageFile = $imageFile;
         $this->escaper = $escaper;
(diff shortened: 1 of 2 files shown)


Block/Menu.php
```diff
@@ -99,8 +99,13 @@ public function __construct(
         $this->templateResolver = $templateResolver;
         $this->imageFile = $imageFile;
         $this->escaper = $escaper;
-        $this->submenuTemplate = $this->getMenuTemplate($this->baseSubmenuTemplate);
         $this->setTemplate($this->getMenuTemplate($this->_template));
+
+        $baseSubmenuTemplate = $this->baseSubmenuTemplate;
```
jochhop, line 104: Please move this into a private method, I think it's too much logic for a constructor

kiber: Done

KamilBalwierz (approved):

Issue #236, opened 2022-02-18
Import from Categories not updating Vue App


michalbiarda:
Steps to reproduce:

1. Create new menu
2. Click "Import from Categories" button and confirm action
3. Wait for server to fetch the data

Result: Vue App is not updated with the results.
Expected result: Vue App is updated with the results.

andrzejborkowski:
@michalbiarda thanks for the PR, we appreciate it! I created an internal ticket and our devs will review it soon.

PR #237, opened 2022-02-18
Fix for Import from Categories not updating Vue App

michalbiarda:
Re #236


RichardOutram:
Have been having the same issue using latest develop branch and 2.4.3.

When using vueApp.$refs.menu.list = list I get
Screenshot 2022-04-19 at 11 05 59

When I use the change vueApp.$children[0].list = list it works correctly.

rjwaberski:
fixed in https://github.com/SnowdogApps/magento2-menu/pull/244

PR #245, opened 2022-05-09
Feature/88914

zuzannaszymanda:


jochhop (approved):

Api/Data/NodeInterface.php
```diff
@@ -262,4 +263,15 @@ public function getAdditionalData();
      * @return $this
      */
     public function setAdditionalData($data);
+
+    /**
+     * @return mixed
```
jochhop, line 268: Wouldn't this return an int type?

Setup/UpgradeSchema.php
```diff
@@ -271,4 +275,24 @@ private function addNodeImageFields(SchemaSetupInterface $setup)
...
     }
+
+    private function addNodeSelectedItemId(SchemaSetupInterface $setup)
+    {
+        $connection = $setup->getConnection();
+        $table = $setup->getTable('snowmenu_node');
+
+        $connection->addColumn(
+            $table,
+            'selected_item_id',
```
jochhop, line 286: Maybe we can use the constant reference from `NodeInterface`?

jochhop (changes requested):

Api/Data/NodeInterface.php
Varmeusz: Thanks for spotting this, changes this with https://github.com/SnowdogApps/magento2-menu/pull/245/commits/1f9444c87ab64da5cfdf3232c9cf092e3df660ba

jochhop (approved):

KamilBalwierz (approved):

PR #268, opened 2023-03-25
Implement product object

marcobolognesi:
The function getCurrentProduct() doesn't work.

https://github.com/SnowdogApps/magento2-menu/blob/d5532c14fc7516791d101991342f52640d63d6dd/Block/NodeType/Product.php#L99-L105

Maybe because the Magento\Framework\Registry is deprecated, however, I added a simple workaround to be able to get the product object in the product.phtml template.

### DIFF
--- a/Block/NodeType/Product.php
+++ b/Block/NodeType/Product.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Snowdog\Menu\Block\NodeType;
 
+use Magento\Catalog\Model\ProductRepository;
 use Magento\Framework\Registry;
 use Magento\Framework\UrlInterface;
 use Snowdog\Menu\Model\TemplateResolver;
@@ -56,6 +57,11 @@ class Product extends AbstractNode
     protected $productTitles;
 
     /**
+     * @var ProductRepository
+     */
+    protected $productRepository;
+
+    /**
      * @var Registry
      */
     private $coreRegistry;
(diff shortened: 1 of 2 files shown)


jochhop:
Hi @marcobolognesi , thanks a lot for your contribution, could you please specify what Magento version do you use so we can test it? thanks in advance!

marcobolognesi:
Hi @jochhop,
the version is 2.4.5-p1
Thank you

jochhop:
Thank you @marcobolognesi ! we'll check it and give you a summary of the tests results soon!

jochhop:
Hello again @marcobolognesi unfortuntally we cannot accept the proposed changes, the reason is that, loading the full project in each node is not performant (that's why we use resource model to populate the needed product data here: https://github.com/SnowdogApps/magento2-menu/blob/develop/Model/ResourceModel/NodeType/Product.php, eg.: `fetchImageData` and `fetchTitleData` methods).

What we propose is to add a new method in `Snowdog\Menu\Model\ResourceModel\NodeType\Product` class to allow passing what attributes should be loaded (similar method to `fetchImageData` or `fetchTitleData` but accepting an array of attribute codes as parameter, this will also allow removing `fetchImageData` or `fetchTitleData` since the new method will do similar stuff)

Would you like to try adding something like above? or would you prefer we do so?

Thanks in advance and have a great day!

marcobolognesi:
I did it because I need the product object to get a custom attribute and also because your method getProductTitle($nodeId) return an empty value
https://github.com/SnowdogApps/magento2-menu/blob/d5532c14fc7516791d101991342f52640d63d6dd/Block/NodeType/Product.php#L284-L287

I close this PR and I open a new one to fix your method.

PR #271, opened 2023-04-12
#99314 add error skipping mechanism for some exceptions

adamwaclawczyk:
Internal ticket DEV-99314

### DIFF
--- a/Model/ImportExport/Processor/Import.php
+++ b/Model/ImportExport/Processor/Import.php
@@ -82,7 +82,7 @@ class Import
     /**
      * @throws ValidationAggregateError
      */
-    private function validateData(array $data): void
+    private function validateData(array &$data): void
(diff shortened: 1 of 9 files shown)


Model/ImportExport/Processor/Import.php
```diff
@@ -71,8 +90,27 @@ private function validateData(array $data): void
             $this->nodeProcessor->validateImportData($data[ExtendedFields::NODES]);
         }
 
-        if ($this->validationAggregateError->getErrors()) {
+        if (empty($this->scopeConfig->getValue('snowmenu/import/strip_invalid_nodes')
```
adamwaclawczyk, line 93: By default == 1, can be changed by adding it to app/etc/env.php or we can add system.xml entry

PR #281, opened 2023-05-17
#100331 Added command to detect invalid menu nodes and remove if needed

jochhop:
It adds a CLI command to detect invalid nodes and remove them if needed

### DIFF
--- a/Console/Command/NodesValidatorCommand.php
+++ b/Console/Command/NodesValidatorCommand.php
@@ -22,6 +22,9 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 
+/**
+ * SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ */
 class NodesValidatorCommand extends Command
 {
     private MenuRepositoryInterface $menuRepository;
@@ -67,31 +70,14 @@ class NodesValidatorCommand extends Command
 
     public function execute(InputInterface $input, OutputInterface $output)
     {
-        try {
-            $this->state->getAreaCode();
-        } catch (LocalizedException $e) {
-            $this->state->setAreaCode(Area::AREA_ADMINHTML);
-        }
+        $this->setAreaCode();
 
         $invalidNodeIds = [];
 
         foreach ($this->getAllMenus() as $menu) {
             $output->writeln(PHP_EOL . '<info>... Validating nodes for menu with ID: ' . $menu->getMenuId(). ' </info>');
             $nodes = $this->nodeRepository->getByMenu($menu->getMenuId());
-            $menuInvalidNodes = [];
-
-            /** @var NodeInterface $node */
-            foreach ($nodes as $node) {
-                $this->validator->validate([$node->getNodeId() => $node->getData()]);
-
-                if (!empty($this->validationAggregateError->getErrors())) {
-                    foreach ($this->validationAggregateError->getErrorMessages() as $errorMessage) {
-                        $menuInvalidNodes[$node->getNodeId()][] = $errorMessage;
-                    }
-
-                    $this->validationAggregateError->flush();
-                }
-            }
+            $menuInvalidNodes = $this->getInvalidNodes($nodes);
 
             if (empty($menuInvalidNodes)) {
                 $output->writeln('<info>No invalid nodes encountered</info>');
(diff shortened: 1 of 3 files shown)


Console/Command/NodesValidatorCommand.php
```diff
@@ -0,0 +1,203 @@
...
+            . PHP_EOL
+            . ' Additionally, it will ask to remove the detected invalid node if needed.';
+
+        $this->setName('snowdog-menu:validate-nodes')
+            ->setDescription($description);
+
+        parent::configure();
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output)
```
adamwaclawczyk, line 71: What do you think about wrapping whole logic in a separate class?

Console/Command/NodesValidatorCommand.php
```diff
@@ -0,0 +1,203 @@
...
+            ->getItems();
+    }
+
+    private function getInvalidNodes(array $nodes): array
+    {
+        $invalidNodes = [];
+
+        /** @var NodeInterface $node */
+        foreach ($nodes as $node) {
+            $this->validator->validate([$node->getNodeId() => $node->getData()]);
```
adamwaclawczyk, line 143: Import validator has `ZERO_BASED_ARRAY_INDEXING_KEY_ADDENED` which makes the later output contain wrong node ids, using `disableNodeIdAddend` should help avoid that

It still deletes nodes using the proper node ids, fixing this would address the wrong node ids under `Invalid nodes encountered for menu ID`

Console/Command/NodesValidatorCommand.php
```diff
@@ -0,0 +1,203 @@
...
+                foreach ($errorMessages as $errorMessage) {
+                    $output->writeln("<error>- $errorMessage</error>");
+                }
+            }
+
+            $invalidNodeIds = array_merge($invalidNodeIds, array_keys($menuInvalidNodes));
+        }
+
+        if (empty($invalidNodeIds) || !$input->isInteractive()) {
+            return;
```
adamwaclawczyk, line 100: Please check this exception message
`Return value of "Snowdog\Menu\Console\Command\NodesValidatorCommand\Interceptor::execute()" must be of the type int, "null" returned.`

Console/Command/NodesValidatorCommand.php
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        $helper = $this->getHelper('question');
+
+        if (!$helper->ask($input, $output, $this->getConfirmationQuestion())) {
+            $output->writeln(PHP_EOL . '<info>Invalid nodes were NOT removed</info>');
+            return;
```
adamwaclawczyk, line 107: Please check this exception message
`Return value of "Snowdog\Menu\Console\Command\NodesValidatorCommand\Interceptor::execute()" must be of the type int, "null" returned.`

Console/Command/NodesValidatorCommand.php
```diff
@@ -0,0 +1,203 @@
...
+                '<error>Invalid nodes encountered for menu ID: ' . $menu->getMenuId() . '</error>'
+            );
+
+            foreach ($menuInvalidNodes as $errorMessages) {
+                foreach ($errorMessages as $errorMessage) {
+                    $output->writeln("<error>- $errorMessage</error>");
+                }
+            }
+
+            $invalidNodeIds = array_merge($invalidNodeIds, array_keys($menuInvalidNodes));
```
adamwaclawczyk, line 96: What do you think about changing this? This breaks rule `Magento2.Performance.ForeachArrayMerge.ForeachArrayMerge`

emastyle: @adamwaclawczyk Coding standard CI on the repo is set at level 10 of severity, so this rule is not intercepted (which is at level of 7).
Actually the markeplace rejects extension for only severity 10 (if we want to follow that ...rule). Anyway we can change the level to 7 in the menu if we want to check just General code issues. https://github.com/magento/magento-coding-standard/wiki/Magento-Marketplace-Extensions-Verification

Console/Command/NodesValidatorCommand.php
jochhop: I think we can, but also I think it's not needed, its logic is not super complex and also useful outputs are shown while command is being executed (which could make more complex to be abstracted)

Anyways if you have a clear idea of how to improve it, please let me know!

Console/Command/NodesValidatorCommand.php
```diff
@@ -140,7 +146,10 @@ private function getInvalidNodes(array $nodes): array
 
         /** @var NodeInterface $node */
         foreach ($nodes as $node) {
-            $this->validator->validate([$node->getNodeId() => $node->getData()]);
+            $this->validator->validate(
+                [$node->getNodeId() => $node->getData()],
+                $this->treeTrace
```
adamwaclawczyk, line 151: Not sure if passing this is required, can you double check?

adamwaclawczyk (changes requested):

Console/Command/NodesValidatorCommand.php
jochhop: it seems to be not needed, thanks!

adamwaclawczyk (approved):

PR #293, opened 2023-09-07
Feature/102193

dawiddczaja:
This feature allows you to use the same custom template for several menus. To do this, when declaring the menu in the xml file, add the parameter "menu_template_dir" It should look like this:

```
<block name="menu-1" class="Snowdog\Menu\Block\Menu">
                <arguments>
                    <argument name="menu" xsi:type="string">menu-1</argument>
                    <argument name="menu_template_dir" xsi:type="string">custom-menu</argument>
                </arguments>
</block>
```


andrzejborkowski:
@adamwaclawczyk I see that related ticket has been rejected, so I guess we can close/remove this PR?

PR #295, opened 2023-09-13
#102271 Adding image size (width and height) to nodes

jochhop:


adamwaclawczyk:
https://github.com/SnowdogApps/magento2-menu/blob/c8b9bcda1cdb78065dedba5ebf8d39518bee70d2/Block/Menu.php#L387
Previous similar PRs that added new fields to nodes, added node info to node blocks inside above method

adamwaclawczyk (approved):

PR #298, opened 2023-10-17
#94738 add customer groups to nodes

adamwaclawczyk:
This adds customer groups field to nodes to add control over which nodes are later rendered on frontend


Model/Menu/Node/Image/File.php
```diff
@@ -123,7 +123,11 @@ public function getImageSize(string $file): array
         $mediaDirectory = $this->filesystem->getDirectoryRead(DirectoryList::MEDIA);
         $fileFullPath = $mediaDirectory->getAbsolutePath(self::PATH . $file);
 
-        return getimagesize($fileFullPath);
+        try {
+            return getimagesize($fileFullPath);
+        } catch (Exception $e) {
+            return [];
```
KamilBalwierz, line 129: what is the point of returning empty array as code calling this is expecting array with information - throw proper exception to be handled in application logic

Model/ResourceModel/Menu/Node.php
```diff
@@ -13,6 +13,29 @@ protected function _construct()
...
     }
 
+    protected function _afterSave(\Magento\Framework\Model\AbstractModel $object)
+    {
+        $connection = $this->getConnection();
+        $connection->delete('snowmenu_customer', ['node_id = ?' => $object->getNodeId()]);
+
+        $nodeCustomerGroups = $object->getData('customer_groups');
+        if ($nodeCustomerGroups && is_string($nodeCustomerGroups)) {
+            $nodeCustomerGroups = json_decode($nodeCustomerGroups);
```
KamilBalwierz, line 23: to serialize this we needed SerializerInterface, now to deserialize just using json_decode - what if serializerIntegrace changes implementation once again ?

PR #299, opened 2023-10-19
Using escapeUrl on block is deprecated

mage2pratik:


jochhop:
Hi @mage2pratik , thank you for your contribution, however we think about leaving the templates as they are for now, since menu module supports older Magento versions, and adding these changes will drop the support prior Magento 2.4 versions.

Could you please verify this argument? 

Thank you in advance,

Jose

mage2pratik:
Hi @jochhop 

Yes, from Magento version 2.4.* $block->escapeUrl is deprecated in templates and in blocks.

jochhop:
Thank you @mage2pratik for now we'll not accept the changes as we want menu module to be compatible with older magento versions (prior 2.4)

So we'll close the PR for now

Thank you a lot for your contribution!

Issue #302, opened 2023-11-16
Node types: wrapper and custom_url have empty url_key in GraphQl context


pmicnasCreatuity:
When you set node with type custom_url or wrapper, the url_key is not set in the graphQl query.

![image-20231114-131853](https://github.com/SnowdogApps/magento2-menu/assets/131661371/1de183ef-d8f4-49f1-9914-c49ccfab3cb0)
![image-20231114-131052](https://github.com/SnowdogApps/magento2-menu/assets/131661371/051ecb72-4714-439a-ade9-178184d489c5)

Example query for the retrieving menu:

`query MenuGetByIdentifier {
  snowdogMenus(identifiers: "menu-id") {
    items {
      ...MenusItem
    }
  }
}
    fragment MenuItem on SnowdogMenuNodeInterface {
  node_id
  parent_id
  title
  url_key
  url_path
  type
  ... on SnowdogMenuCustomUrlNode {
    target
  }
  ... on SnowdogMenuNodeImageFieldInterface {
    image
    image_alt_text
  }
}
fragment MenusItem on SnowdogMenu {
  menu_id
  identifier
  nodes {
    items {
      ...MenuItem
    }
  }
}`

jochhop:
@pmicnasCreatuity thank you for openeing the issue, we have fixed it in the related PR https://github.com/SnowdogApps/magento2-menu/pull/303 and merged to develop branch, the changes will be available in the next menu release.

About wrapper node type, `url_key` is null because that node type does cannot have any content. Please see more details here: https://github.com/SnowdogApps/magento2-menu/wiki/Node-types

Have a nice day!

PR #303, opened 2023-11-21
#103670 Fixed url_key equal to null with custom_url node type

jochhop:
Fixes https://github.com/SnowdogApps/magento2-menu/issues/302


adamwaclawczyk (approved):
Tested using #302 query, with this change `url_key` is now set.

PR #309, opened 2024-01-17
#104299 - Updated snowdog-menu block id

maartynaa:


Model/MenuRepository.php
```diff
@@ -40,8 +43,10 @@ public function __construct(
         MenuFactory $menuFactory,
         CollectionFactory $menuCollectionFactory,
         MenuSearchResultsInterfaceFactory $menuSearchResults,
+        MenuInterface $menu,
```
KamilBalwierz, line 46: I am not a fan of injecting `MenuInterface` directly as it is data object - especially in this context where we also have `MenuFactory` (but this is not true for 2 other instances).
Maybe this should be some kind of helper to provide link fields?

view/adminhtml/templates/menu/nodes.phtml
```diff
@@ -46,6 +46,7 @@ $blockHtmlId = $block->getBlockHtmlId();
                         "imageDeleteUrl"   : "<?= $block->getImageDeleteUrl() ?>",
                         "imageUploadFileId": "<?= $block->getImageUploadFileId() ?>",
                         "lazyMinItemsCount": "1000",
+                        "serializedNodesElementName": "serialized_nodes",
```
ofrankowska, line 49: TODO: pass the correct value from backend (either "serialized_nodes" or "snowmenu_staging_serialized_nodes")

PR #325, opened 2024-05-10
Auto-populate node name field 

ofrankowska:
It would be great to auto-populate node name field after user selects category, category child, CMS page, CMS block or product. Users can edit this name after it has been auto-populated.

image


itaymesh:
Great feature. Do you mind merge it?

andrzejborkowski:
yes, we are still working on it, will be added to one of the next releases

PR #334, opened 2024-08-11
add support for array-data in content

rommelfreddy:
with this PR it is possible to store arrays within the content-field of the Node-Entity. 

This is useful if it is required to store more information, than a single string. 

this helps developer to have structured data, which has not be converted from and to json manually.

### DIFF
--- a/Model/Menu/Node.php
+++ b/Model/Menu/Node.php
@@ -7,6 +7,7 @@ use Magento\Framework\Model\AbstractModel;
 use Magento\Framework\Model\Context;
 use Magento\Framework\Model\ResourceModel\AbstractResource as AbstractResource;
 use Magento\Framework\Registry;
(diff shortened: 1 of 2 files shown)


Model/Menu/Node.php
```diff
@@ -97,14 +98,24 @@ public function setType($type)
      */
     public function getContent()
     {
-        return $this->_getData(NodeInterface::CONTENT);
+        $content =  $this->_getData(NodeInterface::CONTENT);
+
+        if (strpos($content, '{') === 0 && strpos($content, '}') === strlen($content) - 1) {
```
adamwaclawczyk, line 103: `$content` is not always a string, when `_getData()` returns null this will throw an error

adamwaclawczyk (changes requested):

PR #345, opened 2024-10-07
Add Resource Connection to __construct so that the connection table name comes through with any prefixes

antbates91:
Resolves: https://github.com/SnowdogApps/magento2-menu/issues/346

Add Resource Connection to __construct so that the connection table name comes through with any prefixes

Currently, when installed on an instance with a database prefix, the node saving fails with table 'databasename.snowmenu_customer' not found.

This is because `$this->getConnection();` uses a resource that has been instantiated without the concept of Magento database prefixes. 

**Replication Steps**
1. Install Magento with database prefix.
2. Create Menu in admin panel.
3. Add Node type of custom url and fill out information
4. Hit Save.
5. Error `table 'databasename.snowmenu_customer' not found.` is show.


andrzejborkowski:
@antbates91 thanks for your contribution, I created an internal ticket SMM-25, once our dev team will review the code we can add it to the next release.

adamwaclawczyk:
Merged and released in `2.27.2`

PR #352, opened 2024-11-28
SMM-8 Import categories by store view

maartynaa:


Ui/Component/Menu/Form/Element/Categories.php
```diff
@@ -52,15 +76,15 @@ public function toArray(): array
      * @return array
      * @throws LocalizedException
      */
-    private function retrieveCategories(int $storeId = 0, $toOptionArray = true): array
+    private function retrieveCategories(string $storeId = '0', $toOptionArray = true): array
```
adamwaclawczyk, line 79: IMO if possible we should keep the int method declaration isntead

Plugin/Model/ResourceModel/Category/TreePlugin.php
```diff
@@ -0,0 +1,33 @@
...
+    public function beforeAddCollectionData(Tree $subject, $collection = null): array
+    {
+        $postData = $this->request->getPost();
+        $storeId = $postData['store_id'];
+
+        $collection->setProductStoreId(
+            $storeId
+        )->setStoreId(
+            $storeId
+        );
```
adamwaclawczyk, line 29: Maybe we should skip `setProductStoreId` and `setStoreId` if the post param is missing?

Issue #347, opened 2024-10-29
Primary key for 'snowmenu_customer' table


Footshop-Andy:
Hi,

The Adobe Commerce 'Site-Wide Analysis Tool' (SWAT) report recommends a primary key for the 'snowmenu_customer' table.

Do you think one can be implemented in a future release? Thanks!

andrzejborkowski:
hello @Footshop-Andy I created an internal ticket SMM-28 and our dev team will check it soon

PR #354, opened 2024-12-18
SMM-28 add primary key per adobe recommendations

adamwaclawczyk:


adamwaclawczyk:
Fixed https://github.com/SnowdogApps/magento2-menu/issues/347

PR #364, opened 2025-04-17
Shorten long NODE cache tag

ilnytskyi:
In cases when menu contains many nodes, it adds tags to `'x-magento-tags`  header.
Causing errors related to max header size allowed in different services like Varnish or Nodejs

Shortening default tag length will lower the impact.
image

Also can be addressed by increasing header size on Varnish.
Or nodejs, but in some versions it's reported to not working correctly like - https://github.com/nodejs/node/issues/47246

### DIFF
--- a/Model/Menu/Node.php
+++ b/Model/Menu/Node.php
@@ -12,7 +12,7 @@ use Snowdog\Menu\Api\Data\NodeInterface;
 
 class Node extends AbstractModel implements NodeInterface, IdentityInterface
 {
-    const CACHE_TAG = 'snowdog_menu_node';
+    const CACHE_TAG = 'smn';
 
     /**
      * @var SerializerInterface


ilnytskyi:
Another things to consider. 
Do we even need the individual nodes tags?
Can't we just invalidate whole menu cache when one of the nodes changes?

adamwaclawczyk:
Hi @ilnytskyi and thanks for your contribution :)

Changes look ok to me and I think it should not impact anything on Luma/Hyva stores.

I checked some stores and I do not see the `snowdog_menu_node` on any storefront pages - can you share where(e.g. which url) the node cache tags are attached?

>  Do we even need the individual nodes tags?
> Can't we just invalidate whole menu cache when one of the nodes changes?

That's a valid point, I'll raise this internally but feel free to contribute

adamwaclawczyk (approved):

ilnytskyi:
@adamwaclawczyk the tags are added when menu is loaded via graphlq using GET method.

Probably because of this Identity class 
https://github.com/SnowdogApps/magento2-menu/blob/develop/Model/GraphQl/Resolver/Node/Identity.php

PR #366, opened 2025-04-23
Explicitly mark nullable parameters for PHP 8.4 compatibility

### LINKED ISSUE #365: Module doesn't compile in PHP 8.4 due to implicit nullable parameters
PHP 8.4 has deprecated the ability to implicitly treat parameters as nullable and they now require to be explicitly marked as nullable. At present, this issue is preventing setup:di:compile from completing successfully.

Image

I will open a pull request shortly to fix this issue.

samcleathero:
Fixes #365.

### DIFF
--- a/Console/Command/NodesValidatorCommand.php
+++ b/Console/Command/NodesValidatorCommand.php
@@ -44,7 +44,7 @@ class NodesValidatorCommand extends Command
         ValidationAggregateError $validationAggregateError,
         State $state,
         TreeTrace $treeTrace,
-        string $name = null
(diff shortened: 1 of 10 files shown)


adamwaclawczyk (approved):

adamwaclawczyk:
Nullable types were introduced in php7.1 https://www.php.net/manual/en/migration71.new-features.php
I don't think we're supporting versions below 7.4 and we are planning to deprecate support for unsupported php versions: https://www.php.net/supported-versions
So imo we're good to merge

PR #371, opened 2025-07-04
Fixed issue where assigning customer group to a category caused an error when trying to duplicate menu

Riyaprajapat:
**Issue Description:**

When customer groups are assigned to a category in a Menu, attempting to duplicate that menu results an error. As a result, the menu cannot be duplicated successfully.

**Screenshot-1:**
image

**Screenshot-2:**
image

**Screenshot-3:**
image

**Steps to Reproduce:**

1. Take a clean shop.
2. Install snowdog/module-menu in the shop.
3. Go to Stores → Configuration → SNOWDOG → Menu → General and set Customer Groups Enabled to Yes. (See Screenshot-1)
4. Create menu, assign customer groups to a category, and save the menu. (See Screenshot-2)
5. Duplicate that menu and observe result( See Screenshot-3).

**Expected Results:**

The menu should duplicate successfully, even if customer groups are assigned to categories.

**Actual Results:**

An error occurs during the duplication process, and the menu cannot be duplicated.

**Additional Information:**

File path: snowdog/module-menu/Service/Menu/Cloner.php

When cloning the menu, the code at line 119 uses $nodeClone->setData($node->getData()); to copy all data. However, in the original menu, the customer_groups field is saved as a JSON string like ["0", "1", "2", "3"], but after cloning, it becomes a plain array or comma-separated values like 0,1,2,3. This difference in format causes the error “Unable to unserialize value. Error: Syntax error” when saving the cloned menu. To fix this, the customer_groups value should be converted back to a JSON string before saving the cloned node.

image

### DIFF
--- a/Service/Menu/Cloner.php
+++ b/Service/Menu/Cloner.php
@@ -103,8 +103,12 @@ class Cloner
 
             foreach ($this->menuNodes->getList($menu) as $node) {
                 $nodeClone = $this->nodeFactory->create();
+                $data = $node->getData();
 
-                $nodeClone->setData($node->getData());
+                if (isset($data['customer_groups'])) {
+                    $data['customer_groups'] = json_encode((array)$data['customer_groups']);
+                }
+                $nodeClone->setData($data);
                 $nodeClone->setId(null);
                 $nodeClone->setMenuId($menuCloneId);
 


Service/Menu/Cloner.php
```diff
@@ -103,8 +103,12 @@ public function clone(MenuInterface $menu): MenuInterface
 
             foreach ($this->menuNodes->getList($menu) as $node) {
                 $nodeClone = $this->nodeFactory->create();
+                $data = $node->getData();
 
-                $nodeClone->setData($node->getData());
+                if (isset($data['customer_groups'])) {
+                    $data['customer_groups'] = json_encode((array)$data['customer_groups']);
```
adamwaclawczyk, line 109: In code where `customerGroups` is serialized, unserialized we are using `Magento\Framework\Serialize\SerializerInterface`

Let's keep this uniform across this module

adamwaclawczyk (changes requested):

Issue #376, opened 2025-09-22
Relative custom URL breaks when Magento store code is included in URLs


Kashyap1999:
### **Description:**
When adding a Custom URL menu item, internal links are not working as expected.

If you enter an internal path like ‘/contact’ or ‘contact’ (as per screenshot-1), the generated menu link does not include the store’s base URL (see screenshot-2). This causes broken links when Magento is configured to include store codes in URLs.

For example, if you have two stores:
    • **Store A:** https://example.com/ (default)
    • **Store B:** https://example.com/storeB
      
Adding a Custom URL with ‘/contact’ will always render as ‘/contact’ instead of including the correct store code.
    • On Store A → resolves as https://example.com/contact (works)
    • On Store B → resolves as https://example.com/contact instead of **https://example.com/storeB/contact** (broken)

Only full external URLs (e.g. https://google.com) are rendered correctly.

**Screenshot-1**

**![Image](https://github.com/user-attachments/assets/e3ad183b-472c-4e8b-b534-eb782d3ec2a0)**

**Screenshot-2**

**![Image](https://github.com/user-attachments/assets/aec73a2e-7c99-41c0-bac7-878383aa7cf9)**

### **Steps to Reproduce:**
1. Take a multi-store shop with snowdog/module-menu.
2. Go to the Stores > Configuration > General  > Web > Url Options, and enable “Add Store Code to URLs” configuration. (see screenshot-3)
3. Now, Go to Admin > Content > Menus.
4. Add a new Custom URL node.
5. Enter /contact (or contact) as the URL.
6. Save and check the frontend.
7. Inspect the generated HTML.

**Screenshot-3**

**![Image](https://github.com/user-attachments/assets/4526e303-17c4-4b93-aa74-658b6637205c)**

### **Expected Result:**
The internal URL should be resolved with the store base URL, e.g.
`<a href="https://example.com/contact">Contact Us</a>`

As per screenshot-2, category url is generated with the base url same as custom url if not external link than it should be generate with the base url.

### **Actual Result:**
Custom Url is without base url:
`<a href="/contact">Contact Us</a>`

### **Additional Information:**
**Block file:** `snowdog/module-menu/Block/NodeType/CustomUrl.php`
Method `getHtml()` already checks for external vs internal URLs, but it outputs a full `<a>` tag, not just the resolved URL.

**Screenshot-4**

[...]

PR #380, opened 2025-10-07
Fixed relative custom URL breaks when Magento store code is included in URLs

Kashyap1999:
### Description
When adding a Custom URL menu item, internal paths like /contact or contact didn’t include the store’s base URL when “Add Store Code to URLs” was enabled.

This caused links to break on non-default store views — for example, /contact rendered as /contact instead of /storeB/contact.

**Fix:**

- Added logic to resolve internal URLs with the correct store base URL (including store code).
- Updated templates to use the resolved URL method.
- External URLs (e.g., https://google.com) remain unaffected.

**Fixes** #376.

### DIFF
--- a/Block/NodeType/CustomUrl.php
+++ b/Block/NodeType/CustomUrl.php
@@ -112,4 +112,29 @@ HTML;
     {
         return __("Custom Url");
     }
+
+    /**
+    * @param int $nodeId
+    *
+    * @return string
+    */
+    public function getCustomUrl($nodeId)
+    {
+        $node = $this->nodes[$nodeId];
+        $nodeContent = $node->getContent();
+        $url = '';
+
+        if (
+            $nodeContent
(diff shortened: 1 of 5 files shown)


Block/NodeType/CustomUrl.php
```diff
@@ -112,4 +112,29 @@ public function getLabel()
...
     }
+
+    /**
+    * @param int $nodeId
+    *
+    * @return string
+    */
+    public function getCustomUrl($nodeId)
+    {
+        $node = $this->nodes[$nodeId];
```
adamwaclawczyk, line 123: `$this->nodes[$nodeId] ?? null` ?

Kashyap1999: I don’t think this can be null, because this function is called from the custom_url templates when a custom URL is added to the menu and is active. It will always have a nodeId.

PR #396, opened 2026-03-09
Fix GraphQL node loading for duplicate identifiers and add cms block content

sashagfu:
**Summary**
  This PR improves GraphQL support for menu nodes in two areas:

  - adds support for cms_block node content
  - fixes node loading for multi-store setups by loading nodes by resolved menu_id

  **Why**
  In multi-store Magento setups, different menus may use the same identifier across stores. The GraphQL resolver correctly resolves the menu in the current store context first,
  but then loads nodes again by identifier. This may return nodes from other menus that share the same identifier.

  Also, cms_block nodes currently do not return rendered block content in GraphQL.

  **Changes**

  - register cms_block in GraphQL TypeModel
  - load CMS block data through Snowdog\Menu\Model\ResourceModel\NodeType\CmsBlock
  - render CMS block content with FilterProvider before returning it in GraphQL
  - keep url_key unchanged for cms_block nodes
  - replace getByIdentifier() with getByMenu() after resolving the menu entity

  **Validation**
  Validated manually in a real Magento project:

  - cms_block nodes return rendered content in GraphQL
  - GraphQL returns nodes only for the resolved menu in the current store context, even when other stores use the same menu identifier

### DIFF
--- a/Model/GraphQl/Resolver/DataProvider/Node.php
+++ b/Model/GraphQl/Resolver/DataProvider/Node.php
@@ -21,6 +21,7 @@ class Node
     const URL_KEY = 'url_key';
 
     const NODE_TYPE_CUSTOM_URL = 'custom_url';
+    const NODE_TYPE_CMS_BLOCK = 'cms_block';
 
     /**
      * @var NodeRepositoryInterface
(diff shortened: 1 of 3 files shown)


sashatkachukv:
Related Issue https://github.com/SnowdogApps/magento2-menu/issues/397

PR #402, opened 2026-03-30
Manage a single menu instance with different templates for Desktop and Mobile

MaximSR:
## Problem

The Hyva theme uses two separate menu identifiers for desktop and mobile navigation: `hyva-topmenu-desktop` and `hyva-topmenu-mobile`. This requires administrators to configure and maintain two separate menus, even if the content is identical. This leads to duplicated effort and potential inconsistencies between the desktop and mobile menus.

## Solution

This pull request refactors the menu block to decouple the menu's data source from its rendering template. This allows developers to use one menu as a data source while rendering it with a template from another menu identifier.

The solution introduces a new `template-identifier` argument for the menu block and ensures backward compatibility.

The key changes are:

1.  **`Block/Menu.php`**:
    *   The `getMenuTemplate` method is updated to prioritize the new `template-identifier` argument for template resolution.
    *   If the `template-identifier` is not set, the block falls back to using the `menu` argument to find the template. This maintains full backward compatibility for existing menu block implementations.

2.  **`view/frontend/layout/default_hyva.xml`**:
    *   The `topmenu-mobile` can use `hyva-topmenu-desktop` as its data source on demand (via the `menu` argument).
    *   Its presentation is sourced from the `hyva-topmenu-mobile` template (via the `template-identifier` argument).
    *   This allows administrators to manage the menu content in a single place (`hyva-topmenu-desktop`) while still having distinct layouts for desktop and mobile.


andrzejborkowski:
hey @MaximSR,

thank you for your PR. I’ve created an internal ticket (SMM-94), and we’ll review the suggested approach with the dev team to see whether it is acceptable.
From our perspective, it would be better to manage desktop and mobile menus separately. This allows us to optimize navigation for customers on different devices, which has worked well across most of our projects.
A key aspect is that on mobile, top-level menu nodes are clickable without redirecting users to the category page. Instead, they open sub-nodes along with a “View All” link at the top, allowing users to access all products in that category. This behavior is essential for mobile UX and would be difficult to maintain if the menus were unified.
image

Regards,
Andrzej

PR #393, opened 2026-02-01
Fixes #391

codingkoaladev:

Issue #391, opened 2026-01-22
Invalid template file: 'Snowdog_Menu::hyva-topmenu-mobile/menu/custom/sub_menu/3-level.phtml' in module: 'Snowdog_Menu' block's name: 'topmenu-mobile'


codingkoaladev:
Steps to replicate:

Create a desktop menu with a category and some sub categories

<img width="1474" height="379" alt="Image" src="https://github.com/user-attachments/assets/4f5460fe-0569-42de-af78-2201d669f320" />

Clone the menu and assign as the mobile menu

<img width="1430" height="155" alt="Image" src="https://github.com/user-attachments/assets/442cb546-d9f8-420b-9661-32b7ddf0abab" />

Refresh frontend, the error below gets thrown

1 exception(s):
Exception #0 (Magento\Framework\Exception\ValidatorException): Invalid template file: 'Snowdog_Menu::hyva-topmenu-mobile/menu/custom/sub_menu/3-level.phtml' in module: 'Snowdog_Menu' block's name: 'topmenu-mobile'

Exception #0 (Magento\Framework\Exception\ValidatorException): Invalid template file: 'Snowdog_Menu::hyva-topmenu-mobile/menu/custom/sub_menu/3-level.phtml' in module: 'Snowdog_Menu' block's name: 'topmenu-mobile'
<pre>#1 Snowdog\Menu\Block\Menu\Interceptor->fetchView() called at [vendor/mage-os/framework/View/Element/Template.php:293]
#2 Magento\Framework\View\Element\Template->_toHtml() called at [vendor/mage-os/framework/View/Element/AbstractBlock.php:1128]
#3 Magento\Framework\View\Element\AbstractBlock->Magento\Framework\View\Element\{closure}() called at [vendor/mage-os/framework/View/Element/AbstractBlock.php:1132]
#4 Magento\Framework\View\Element\AbstractBlock->_loadCache() called at [vendor/mage-os/framework/View/Element/AbstractBlock.php:676]
#5 Magento\Framework\View\Element\AbstractBlock->toHtml() called at [vendor/snowdog/module-menu/Block/Menu.php:319]
#6 Snowdog\Menu\Block\Menu->renderSubmenu() called at [vendor/snowdog/module-menu/view/frontend/templates/hyva-topmenu-mobile/menu.phtml:124]
#7 include() called at [vendor/mage-os/framework/View/TemplateEngine/Php.php:67]
[...]

Issue #379, opened 2025-10-06
Duplicated menu with changed identifier throws an error if same submenu template does not exist.


manan-patel1:
### **Preconditions (*)**
Snowdog_Menu module version : 2.29.0

### **Description (*)**
When duplicating a menu and changing its identifier in hyva theme, the submenu template from the original menu is copied over. If the new identifier does not have a corresponding submenu template, the frontend breaks and displays an error(Screenshot - 1).

Screenshot-1:
<img width="1899" height="866" alt="Image" src="https://github.com/user-attachments/assets/90301bc5-898c-4f26-a035-f197aa661724" />

### **Steps to reproduce (*)**

1. Setup magento shop with hyva-theme.
2. Install the “snowdog/module-menu” module.
3. Create some categories and add products to it.
4. There are already two identifiers created there hyva-topmenu-desktop and hyva-topmenu-mobile.
5. Create a new menu with identifier hyva-topmenu-desktop, add categories and subcategories(same as parent node and child node), and assign a submenu template(called 3-level) specific to hyva-topmenu-desktop.
6. Duplicate the created menu.
7. Change its identifier to hyva-topmenu-mobile and enable the menu.
8. Refresh cache and check the frontend.
9. It will show an error as shown in screenshot-1.

### **Expected result (*)**
When changing the identifier on a duplicated menu, if the previously assigned submenu template does not exist for the new identifier, the module should gracefully handle the missing template and ideally display an admin warning or validation message to inform the user about the missing submenu template for that menu.

### **Actual result (*)**
The frontend throws an error (see Screenshot 1) and the menu doesn’t render. No validation or fallback mechanism is triggered to handle the missing submenu template.

### **Additional Information (*)**
**File Path**: vendor/snowdog/module-menu/Block/Menu.php

In the mentioned file, the getSubmenuBlock() method retrieves and renders the submenu template. However, when a menu is duplicated, all data — including the submenu template name — is copied from the original menu. If the identifier is later changed and the newly selected identifier does not have the same submenu template, the method still attempts to load the copied template(Screenshot-2). This results in a missing template error on the frontend.

Screenshot-2:
<img width="1336" height="537" alt="Image" src="https://github.com/user-attachments/assets/e3659c20-6bba-4bdb-9adc-a89ef4e8a69c" />

### **Reference (*)**
[...]

PR #416, opened 2026-09-07
Fix duplicated menu with changed identifier throws an error if same submenu template does not exist

manan-patel1:
## Description

Fixes an issue where a duplicated menu can break the frontend when its configured submenu template does not exist for the new menu identifier.

This issue was originally reported in [#379](https://github.com/SnowdogApps/magento2-menu/issues/379) and later reproduced with an `Invalid template file` exception in [#391](https://github.com/SnowdogApps/magento2-menu/issues/391).

### Previous Approach

The initial fix was proposed in [PR #393](https://github.com/SnowdogApps/magento2-menu/pull/393).

The previous implementation added Hyva-specific template validation to `Block/Menu.php`. Since `Block/Menu.php` is shared across themes and should remain theme-agnostic, this approach was not suitable.

### Changes

- Added an `isValidTemplate()` method to `TemplateResolver.php` to validate whether a template is available and valid within the current theme/module layout framework.
- Updated `Menu.php` to validate the custom submenu template before using it.
- If the custom submenu template is not valid for the current menu identifier, the menu falls back to the default submenu template.
- Keeps `Block/Menu.php` theme-agnostic without introducing Hyva-specific logic.

### How the Fix Solves It

When a menu is duplicated and its identifier is changed, the copied submenu template may no longer exist for the new identifier.

The custom submenu template is now validated before it is assigned:

1. The custom submenu template path is constructed using the current menu identifier.
2. `isValidTemplate()` checks whether the template is available and valid.
3. If the template is valid, it is used as before.
4. If the template is not valid, the default submenu template is used instead.

This prevents an invalid submenu template from breaking the frontend while preserving the existing submenu template configuration.

## How to Test

1. Create a new menu with the identifier `hyva-topmenu-desktop`.
2. Add categories and subcategories to the menu and create a parent/child hierarchy.
3. Assign the `3-level` submenu template, ensuring that it is specific to `hyva-topmenu-desktop`.
4. Duplicate the menu and change its identifier to `hyva-topmenu-mobile`.
5. Enable the duplicated menu.
6. Clear the cache and refresh the frontend.

### Expected Result

The frontend should render the menu using the default submenu template instead of throwing an `Invalid template file` exception.

### Related Issues / PRs

- Original issue: [#379](https://github.com/SnowdogApps/magento2-menu/issues/379)
- Referenced issue: [#391](https://github.com/SnowdogApps/magento2-menu/issues/391)
- Previous proposed fix: [#393](https://github.com/SnowdogApps/magento2-menu/pull/393)

### DIFF
--- a/Block/Menu.php
+++ b/Block/Menu.php
@@ -445,9 +445,15 @@ class Menu extends Template implements DataObject\IdentityInterface
     {
         $block = clone $this;
         $submenuTemplate = $parentNode->getSubmenuTemplate();
-        $submenuTemplate = $submenuTemplate
-            ? 'Snowdog_Menu::' . $this->getMenu()->getIdentifier() . "/menu/custom/sub_menu/{$submenuTemplate}.phtml"
-            : $this->submenuTemplate;
+
+        if ($submenuTemplate) {
+            $customSubmenuTemplate = 'Snowdog_Menu::' . $this->getMenu()->getIdentifier() . "/menu/custom/sub_menu/{$submenuTemplate}.phtml";
+            $submenuTemplate = $this->templateResolver->isValidTemplate($block, $customSubmenuTemplate)
+                ? $customSubmenuTemplate
+                : $this->submenuTemplate;
+        } else {
+            $submenuTemplate = $this->submenuTemplate;
+        }
 
         $block->setSubmenuNodes($nodes)
             ->setParentNode($parentNode)
--- a/Model/TemplateResolver.php
+++ b/Model/TemplateResolver.php
@@ -116,6 +116,16 @@ class TemplateResolver
     }
 
     /**
+     * @param Template $block
+     * @param string $template
+     * @return bool
+     */
+    public function isValidTemplate($block, $template)
+    {
+        return (bool) $this->validator->isValid($block->getTemplateFile($template));
+    }
+
+    /**
      * @param string $nodeType
      * @param string $defaultTemplateLabel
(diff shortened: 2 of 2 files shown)


Block/Menu.php
```diff
@@ -445,9 +445,15 @@ private function getSubmenuBlock($nodes, $parentNode, $level = 0)
     {
         $block = clone $this;
         $submenuTemplate = $parentNode->getSubmenuTemplate();
-        $submenuTemplate = $submenuTemplate
-            ? 'Snowdog_Menu::' . $this->getMenu()->getIdentifier() . "/menu/custom/sub_menu/{$submenuTemplate}.phtml"
-            : $this->submenuTemplate;
+
+        if ($submenuTemplate) {
+            $customSubmenuTemplate = 'Snowdog_Menu::' . $this->getMenu()->getIdentifier() . "/menu/custom/sub_menu/{$submenuTemplate}.phtml";
+            $submenuTemplate = $this->templateResolver->isValidTemplate($block, $customSubmenuTemplate)
```
jupiterlewis: @manan-patel1 isValidTemplate does a filesystem call. Since this is running for every submenu item, the validity should be cached. Take a look at Model/TemplateResolver.php getMenuTemplate(). Check if you can use that instead.

jupiterlewis (changes requested):

jupiterlewis (approved):

